PR #7, opened 2021-07-28
update gh-page to show a nice page when opening via browser

cpanato:
update gh-page to show a nice page when opening via browser


sabre1041 (approved):
LGTM

PR #8, opened 2021-08-01
Management of mysql secret

sabre1041:
Several corrections to how mysql secret is managed

* Fix incorrect reference to `mysql.auth.rootPassword`
* Reuse existing secret content upon upgrade

### DIFF
--- a/charts/rekor/templates/_helpers.tpl
+++ b/charts/rekor/templates/_helpers.tpl
@@ -461,3 +461,20 @@ Return the location for file based Attestation storage.
 {{- define "rekor.server.fileAttestationStorage.path" -}}
 {{- print (substr 7 -1 .Values.server.attestation_storage.bucket) -}}
 {{- end -}}
+
+{{/*
+Return a random Secret value or the value of an exising Secret key value
+*/}}
+{{- define "rekor.randomSecret" -}}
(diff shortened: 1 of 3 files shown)


charts/rekor/templates/_helpers.tpl
```diff
@@ -460,4 +460,21 @@ Return the location for file based Attestation storage.
...
 {{- print (substr 7 -1 .Values.server.attestation_storage.bucket) -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{/*
+Return a random Secret value or the value of an exising Secret key value
+*/}}
+{{- define "rekor.randomSecret" -}}
+{{- $randomSecret := (randAlpha 10) }}
```
SantiagoTorres, line 469: I wonder if you'd like to use randAlphaNum or even randAscii so as to increase the character set of the secret (are these secrets only meant to use `A-Za-z`?

SantiagoTorres (changes requested):
Mostly LGTM, minor entropy-related nit, but can be fixed (or just explained without a re-roll)

charts/rekor/templates/_helpers.tpl
sabre1041: Updated

SantiagoTorres (approved):

PR #9, opened 2021-08-24
[wip] feat: add cosigned chart

hectorj2f:
Signed-off-by: Hector Fernandez <hectorf@vmware.com>

This PR adds the helm chart for the installation of a sigstore admission controller for cosign. We are still figuring it out other details such as the image name and the release process under `sigstore/cosign` repo.

It is related to https://github.com/sigstore/cosign/issues/550.

### DIFF
--- a/charts/cosigned/README.md
+++ b/charts/cosigned/README.md
@@ -1,0 +1,90 @@
+# Cosigned Admission Webhook
+
+## Requirements
+* kind (or any other Kubernetes cluster).
+* Helm.
+
+## Run `cosigned`
+
+Cosigned requires `cert-manager` to be pre-configured on the running cluster.
+To install `cert-manager` follow the next steps:
+
+```shell
+helm repo add jetstack https://charts.jetstack.io
+helm repo update
+helm install \
+  cert-manager jetstack/cert-manager \
+  --namespace cert-manager \
+  --create-namespace \
+  --version v1.5.0 \
+  --set installCRDs=true
+```
+
+Once `cert-manager` is installed in your cluster, you can start configuring `cosigned`.
+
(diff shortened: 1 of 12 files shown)


charts/cosigned/README.md
```diff
@@ -0,0 +1,90 @@
...
+  --create-namespace \
+  --version v1.5.0 \
+  --set installCRDs=true
+```
+
+Once `cert-manager` is installed in your cluster, you can start configuring `cosigned`.
+
+Generate a keypair to validate the signatures of the deployed Kubernetes resources and their images:
+
+```shell
```
developer-guy, line 27: we can use the following commands to store key pairs in Kubernetes secret, you can get more detail about it from [here](https://github.com/sigstore/cosign/blob/main/KMS.md#kubernetes-secret).
```shell
# generate a key-pair in Kubernetes Secret
cosign generate-key-pair k8s://[NAMESPACE]/[NAME]
```

so, the user does not have to create a secret manually to store these. 🙏

developer-guy (changes requested):

developer-guy:
Quick note: currently, we are supporting only one secret for storing key pairs but in the case of other key pairs we have to create additional secrets, so we have to provide a way to consume these secrets to load key pairs as @dlorenc did with his ConfigMap feature to store more than one key pair.

WDYT @hectorj2f @dlorenc @Dentrax ?

charts/cosigned/README.md
hectorj2f: This is another thing to improve, i just let it as it was from the original code. But it definitely needs a re-design or re-think this approach so it can works with all the supported options (e.g. KMS, ...)

PR #10, opened 2021-09-02
charts: move cosigned chart from s/cosign to the helm repo

cpanato:
This PR adds the cosigned helm chart that exists in the sigstore/cosign repo to here, which is the place for all Helm Charts for sigstore.

A follow-up PR is to remove and clean cosigned chart and ci infra from cosign repo


charts/cosigned/templates/webhook/webhook_validating.yaml
```diff
@@ -0,0 +1,69 @@
...
+    {{- include "cosigned.labels" . | nindent 4 }}
+    control-plane: {{ template "cosigned.fullname" . }}-webhook
+webhooks:
+  - clientConfig:
+      service:
+        name: {{ template "cosigned.fullname" . }}-webhook
+        namespace: {{ .Release.Namespace }}
+        path: /validations
+    failurePolicy: Ignore
+    #namespaceSelector:
```
sabre1041, line 32: Given the scope of the types of resources that are affected by this policy, there has to be a configurable way to limit the blast radius

sabre1041 (changes requested):
Good initial implementation. made a few suggestions/comments

charts/cosigned/README.md
```diff
@@ -0,0 +1,98 @@
...
+Generate a keypair to validate the signatures of the deployed Kubernetes resources and their images:
+
+```shell
+cosign generate-key-pair
+```
+
+The previous command generates two key files `cosign.key` and `cosign.pub`. Next, create a secret to validate the signatures:
+
+```shell
+kubectl create secret generic mysecret -n cosigned --from-file=cosign.pub=./cosign.pub
```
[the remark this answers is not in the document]
cpanato: added the step to create the namespace

charts/cosigned/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+```shell
+helm repo add sigstore https://sigstore.github.io/cosign/
+
+helm repo update
+
+helm install cosigned -n cosigned sigstore/cosigned --devel --set webhook.secretKeyRef.name=k8s://cosigned/mysecret --create-namespace
+```
+
+We need to add the `--devel` flag because we are still in the development of the chart. This will be removed when we release cosigned `v1.0.0`
```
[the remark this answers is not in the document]
cpanato: we don't have yet, will be available in the next cosign release and then we update the docs and the tag

charts/cosigned/templates/webhook/webhook_validating.yaml
cpanato: did an initial customization

charts/cosigned/README.md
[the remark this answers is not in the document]
cpanato: moving the image to the one we are building in CI for now instead of using the one the Hector built in his own registry

cpanato:
While reviewing the feedback I notice that we made several changes in the cosigned controller and now we are pushing the CI image to the sigstore registry as well so we can use that (we will update to the tagged release when 1.2.0 is out)
But the current chart did not work with the new cosigned changes.

Did the refactor in the chart to make it work, and this is the initial thing, I would suggest any customization or other improvement we can make in follow-ups PRs.


testing locally:

- Installing the chart
```shell
$ helm install cosigned -n cosign-system -f ci/ci-values.yaml  .
NAME: cosigned
LAST DEPLOYED: Sun Sep 12 14:59:14 2021
NAMESPACE: cosign-system
STATUS: deployed
REVISION: 1
TEST SUITE: None

$ k get po
NAME                                READY   STATUS    RESTARTS   AGE
cosigned-webhook-5cd889f454-rpsgr   1/1     Running   0          35s

```

Add a namespace to be watched:
```shell
$ k describe ns test1
Name:         test1
Labels:       cosigned.sigstore.dev/include=true
              kubernetes.io/metadata.name=test1
Annotations:  <none>
Status:       Active

No resource quota.

No LimitRange resource.
```

- Applying a deployment that is not signed
```
$ k apply -f testimage.yaml -n test1
Error from server (BadRequest): error when creating "testimage.yaml": admission webhook "cosigned.sigstore.dev" denied the request: validation failed: invalid image signature: spec.template.spec.containers[0].image
```


- applying a deployment that has an Image signed
```
$ k apply -f testimage-signed.yaml -n test1
deployment.apps/nginx-deployment created
```


/cc @hectorj2f @dlorenc @mattmoor @sabre1041 

and thanks so much for your review and feedback @sabre1041

hectorj2f (approved):
lgtm, and works for me

charts/cosigned/templates/webhook/deployment_webhook.yaml
```diff
@@ -0,0 +1,76 @@
...
+          value: "{{ $value }}"
+{{- end }}
+{{- end }}
+        args:
+        - -secret-name={{ template "cosigned.fullname" . }}-cosign-key
+        {{- range $key, $value := .Values.webhook.extraArgs }}
+        - -{{ $key }}={{ $value }}
+        {{- end }}
+        ports:
+        - containerPort: 8443
```
hectorj2f, line 54: Do we want to set the metrics port here ?
```
        - containerPort: 9090
          name: metrics
          protocol: TCP 
```

cpanato: will fix that

ct.yaml
```diff
@@ -1,3 +1,8 @@
 chart-dirs:
   - charts
 validate-maintainers: false
+remote: origin
+target-branch: main
+chart-repos:
+  - stable=https://charts.helm.sh/stable
```
mrueg, line 7: are those needed? I don't see any dependencies that would require the repos

charts/cosigned/templates/webhook/deployment_webhook.yaml
cpanato: done, thanks!

ct.yaml
cpanato: that is true! thanks for the catch!

mrueg:
Not sure if you want to include this in the first PR as well:  Metrics could be exposed via a service and an optional ServiceMonitor can be added (happy to do a follow-up PR, if desired).

cpanato:
> Not sure if you want to include this in the first PR as well: Metrics could be exposed via a service and an optional ServiceMonitor can be added (happy to do a follow-up PR, if desired).

I think we can do in a follow up

hectorj2f (approved):

hectorj2f:
Should we add more maintainers to this repo? I am not sure if @sabre1041 is alone here.

cpanato:
> Should we add more maintainers to this repo? I am not sure if @sabre1041 is alone here.

agree, but maybe this something that @sabre1041 and @dlorenc can let us know :)

charts/cosigned/templates/webhook/cosign_secret.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    {{- include "cosigned.labels" . | nindent 4 }}
+  name: {{ template "cosigned.fullname" . }}-cosign-key
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+data:
+  cosign.key: {{ .Values.cosign.cosignKey}}
```
dol, line 10: `{{ required "A cosign key is required" .Values.cosign.cosignKey }}`
> see
https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function

This helps checking for values that are required.

An other idea would could be to add a `values.schema.json` => https://helm.sh/docs/topics/charts/#schema-files

cpanato: thanks for that, i forgot to add that we used to have it but I lost in the transition, will update

hectorj2f:
@sabre1041 Could you have a look at this PR 😍  ?

cpanato:
flipped to use the official release: `gcr.io/projectsigstore/cosigned:v1.2.0` 🎉

charts/cosigned/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  cosignKey:
+  cosignPub:
+  cosignPassword:
+
+webhook:
+  name: webhook
+  image:
+    repository: gcr.io/projectsigstore/cosigned
+    # crane digest gcr.io/projectsigstore/cosigned:v1.2.0
+    version: sha256:55640ca4ad7803d0002d97c0200636b37318570702a74bd1f7650e1fda6206ae
```
dol, line 12: To preserve the image tag information the version could be changed to `v1.2.0@sha256:55640ca4ad7803d0002d97c0200636b37318570702a74bd1f7650e1fda6206ae`.

The format `name[:tag|@digest]` is supported by k8s. This admission webhook tool performs this action on pods => https://github.com/google/k8s-digester

Docker and podman also support this notion:
- https://docs.docker.com/engine/reference/commandline/image_pull/
- https://docs.podman.io/en/latest/markdown/podman-pull.1.html

cpanato: agree with you, but we can change that in a follow-up, I did in this way per Andrew's request to match the other chart we have here, but I agree to have the tag on it that will be easier :)

PR #11, opened 2021-09-13
minor reword of the description

axelsimon:
might be nitpicking, but hope it helps

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: rekor
-description: Rekor is an API based server for validation and a transparency log for storage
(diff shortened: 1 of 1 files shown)


sabre1041:
@axelsimon not nitpicking at all as it helps to more clearly describe the functions and capabilities of Rekor. When a change is introduced, such as this, it would require a bump to the chart version. I have a change that I'd like to introduce in the next few days. I can incorporate the changes in this PR then, or if you want to update the chart version in this PR, we can go down that route as well.

sabre1041:
@axelsimon incorporated into #12

axelsimon:
@sabre1041 Sorry for the slow reply, fell through the cracks. Bundling this with your change sounds good, thanks for going ahead and doing so :)

PR #12, opened 2021-09-16
Updates to image values format and JSON Schema for Rekor

sabre1041:
Updates to image values format and JSON Schema for Rekor


charts/rekor/values.yaml
```diff
@@ -12,7 +12,8 @@ redis:
     - "yes"
   name: redis
   image:
-    repository: docker.io/redis
+    registry: docker.io
+    repository: redis
     pullPolicy: IfNotPresent
     version: "5.0.10"
```
hectorj2f, line 18: We should use the digest here.`sha256: 4f226f7d68a9880b3f4952d53d5d57c0f3b0c4941b455f7279a12c63c9004a00`

sabre1041: @hectorj2f Updated to include the value of the digest. `sha256:0a0d563fd6fe5361316dd53f7f0a244656675054302567230e85eb114f683db4` was the value used as its the amd64 digest instead of the arm version you referenced (why dockerhub defaults to arm is another story :) ). 

Thanks again for the suggestion to move over to digests for this image!

charts/rekor/values.yaml
```diff
@@ -12,9 +12,10 @@ redis:
     - "yes"
   name: redis
   image:
-    repository: docker.io/redis
+    registry: docker.io
+    repository: redis
     pullPolicy: IfNotPresent
-    version: "5.0.10"
+    version: "sha256:0a0d563fd6fe5361316dd53f7f0a244656675054302567230e85eb114f683db4"
```
cpanato, line 18: don't we make a comment to say which tag is that? will be useful when updating the image. Or add a comment in how to discovery from which release is that.

same for others.

cpanato (approved):
small non blocking comment

hectorj2f (approved):

charts/rekor/values.yaml
sabre1041: @cpanato updated

cpanato: thanks!

PR #15, opened 2021-09-17
cosigned: add scripts to expose metrics to prometheus

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

#### Ticket Link

Fixes

#### Release Note

```release-note
cosigned: Add flag to expose metrics to prometheus
```
This PR adds the configmap, service and service monitor needed to expose the metrics in Prometheus.

### DIFF
--- a/charts/cosigned/Chart.yaml
+++ b/charts/cosigned/Chart.yaml
@@ -5,7 +5,7 @@ sources:
   - https://github.com/sigstore/cosign
 name: cosigned
 type: application
-version: v0.0.3-dev
+version: v0.0.4
 appVersion: v1.2.0
 maintainers:
   - name: dlorenc
--- a/charts/cosigned/templates/webhook/configmap.yaml
+++ b/charts/cosigned/templates/webhook/configmap.yaml
@@ -32,7 +32,7 @@ data:
   # Changes are be picked up immediately.
   loglevel.controller: "info"
   loglevel.webhook: "info"
-{{- if and .Values.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
+{{- if .Values.serviceMonitor.enabled }}
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/charts/cosigned/templates/webhook/configmap.yaml
+++ b/charts/cosigned/templates/webhook/configmap.yaml
@@ -32,3 +32,17 @@ data:
   # Changes are be picked up immediately.
   loglevel.controller: "info"
   loglevel.webhook: "info"
+{{- if and .Values.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    {{- include "cosigned.labels" . | nindent 4 }}
+    control-plane: {{ template "cosigned.fullname" . }}-webhook
+  name: {{ template "cosigned.fullname" . }}-webhook-observability
+  namespace: {{ .Release.Namespace }}
+data:
+  metrics.backend-destination: prometheus
+  metrics.request-metrics-backend-destination: prometheus
+{{- end }}
--- a/charts/cosigned/templates/webhook/service_monitor_webhook.yaml
+++ b/charts/cosigned/templates/webhook/service_monitor_webhook.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
+{{- if .Values.serviceMonitor.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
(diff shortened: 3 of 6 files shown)


mattmoor (approved):
👍

hectorj2f:
@sabre1041 Could you have a look ?

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
```
dol, line 1: I would generally avoid `Capabilities` and `lookup` in helm charts. Some tools like ArgoCD, helm diff ( https://github.com/databus23/helm-diff/issues/165 ) use `helm template` instead of `helm install/upgrade`. `helm template` has no connection to the Kubernetes API server and therefor the condition will never be true.
I would let the apply of the ServiceMonitor fail if the resource is unknown to the cluster.
The feature flag `serviceMonitor.enabled` should be enough.

sabre1041: agree with this comment. the lack of support through GitOps tools, like argocd

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: ServiceMonitor
+metadata:
+  name: {{ template "cosigned.fullname" . }}-webhook-monitor
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cosigned.labels" . | nindent 4 }}
+    control-plane: {{ template "cosigned.fullname" . }}-webhook
+spec:
+  endpoints:
+    - path: /metrics
```
sabre1041, line 12: Any reason why we might want to parameterize the metrics path to a value?

charts/cosigned/Chart.yaml
```diff
@@ -5,7 +5,7 @@ sources:
   - https://github.com/sigstore/cosign
 name: cosigned
 type: application
-version: v0.0.3-dev
+version: v0.0.4-dev
```
cpanato, line 8: we can remove the -dev

charts/cosigned/templates/webhook/configmap.yaml
```diff
@@ -32,3 +32,17 @@ data:
...
+metadata:
+  labels:
+    {{- include "cosigned.labels" . | nindent 4 }}
+    control-plane: {{ template "cosigned.fullname" . }}-webhook
+  name: {{ template "cosigned.fullname" . }}-webhook-observability
+  namespace: {{ .Release.Namespace }}
+data:
+  metrics.backend-destination: prometheus
+  metrics.request-metrics-backend-destination: prometheus
+{{- end }}
```
cpanato, line 48: nit: new line

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
cpanato: no strong opinions but makes sense 
+1

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
```diff
@@ -0,0 +1,17 @@
...
+      port: metrics
+  selector:
+    matchLabels:
+      control-plane: {{ template "cosigned.fullname" . }}-webhook
+{{- end }}
```
cpanato, line 17: nit: new line

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
hectorj2f: Why don't we want it ? It is the default path value and, as far as I remember, it is not configurable.

cpanato (approved):

charts/cosigned/templates/webhook/configmap.yaml
```diff
@@ -32,3 +32,17 @@ data:
   # Changes are be picked up immediately.
   loglevel.controller: "info"
   loglevel.webhook: "info"
+{{- if and .Values.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
```
dol, line 35: Same as https://github.com/sigstore/helm-charts/pull/15/commits/5988aef0f48881338b5d4c655a130a93c36546ee#r717480526

hectorj2f: Done! Good catch

charts/cosigned/templates/webhook/service_monitor_webhook.yaml
cpanato: I also guess that is not configurable, would suggest if that is configurable we can update this in a follow-up.

waiting your comment @sabre1041

sabre1041: Lets get it in

PR #16, opened 2021-10-01
Specify paths to trigger workflows

sabre1041:
Trigger workflows only when changes occur to `charts` directory


cpanato (approved):
thanks!

PR #32, opened 2021-11-08
rekor: Set storageClass types to string in schema

### LINKED ISSUE #31: rekor: cannot specify storage classes due to schema error
**Description**

For the Rekor chart, neither of the two `storageClass` parameters can be set to non-null values due to the schema specifying the type as `null`.

The two values in question are:
1. `mysql.persistence.storageClass`
2. `server.attestation_storage.persistence.storageClass`

These can be fixed by modifying the schema file to set both types as `string`.

syphr42:
#### Summary

This change allows the user to specify non-null values for storageClass for both MySQL and Rekor. Without this change, a schema validation error is thrown when any non-null value is supplied for either of these storageClass variables.

#### Ticket Link

Fixes #31

#### Release Note

```release-note
Fixed validation of non-default (non-null) storageClass values for the Rekor chart
```

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.4
+version: 0.2.5
 appVersion: 0.3.0
 
 keywords:
--- a/charts/rekor/values.schema.json
+++ b/charts/rekor/values.schema.json
@@ -1700,7 +1700,10 @@
                         },
                         "storageClass": {
                             "$id": "#/properties/mysql/properties/persistence/properties/storageClass",
-                            "type": "null",
+                            "type": [
+                                "string",
+                                "null"
+                            ],
                             "title": "The storageClass schema",
                             "description": "An explanation about the purpose of this instance.",
                             "default": null,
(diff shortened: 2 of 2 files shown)


cpanato (changes requested):
please bump the chart version

cpanato (approved):

syphr42:
I noticed the linter failures. It seems like `null` and `string` are incompatible types *and* `null` is not valid as a `string`. I tested locally replacing all the default `null` values of `storageClass` with an empty string (`""`) and that appears to work. I'm not very familiar with the schema validations component of Helm, but if someone could confirm that my understanding makes sense, I'm happy to update this PR with those changes.

However, there is a potential user impact here. After changing the `null` to `""`, if someone were explicitly passing `null` for `storageClass` in their values.yaml, it would fail after taking this chart update due to schema validation. Is that enough to bump the major version number on the chart or not a likely issue?

sabre1041:
@syphr42 @cpanato As an FYI, I have this queued up to work on Friday evening

sabre1041:
@syphr42 JSON Schema types can support multiple values. If you specify both `string` and `null` as follows, validation should pass. 

```
"type": [
    "string",
    "null"
],
```

Also, be sure to rebase from `main` as https://github.com/sigstore/helm-charts/pull/27 was recently merged

syphr42:
@sabre1041 This should be good now. The linter was successful on a local run and deployment to a test k8s cluster with storageClass values set to something other than `null` works.

sabre1041 (approved):
LGTM

PR #39, opened 2021-11-16
rekor: refactor service wait init containers

syphr42:
#### Summary

The goal of this change is to upgrade curl to the current stable release. However, since the current stable release of curl behaves differently with respect to non-HTTP servers, a different solution was necessary for determining availability of the database (MySQL and MariaDB do not use HTTP).

This resulted in three changes:

1. Change the curl image to docker.io/curlimages/curl:7.79.1
2. Modify the test for database server availability to use netcat instead of curl
3. Refactor the initContainerImage section of the values.yaml to support multiple init containers

#### Ticket Link

n/a

#### Release Note

```release-note
rekor: refactored service wait init containers which includes a change to the structure of the initContainerImage section of values.yaml
```

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.5
+version: 0.2.6
 appVersion: 0.3.0
 
 keywords:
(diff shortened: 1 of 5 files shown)


cpanato (approved):
thanks so much for the update

looks good to me

sabre1041 (approved):
LGTM

PR #48, opened 2021-12-18
Support for signed released

sabre1041:
Support for signing Helm releases and publishing to Rekor

Requires `GPG_PRIVATE_KEY` and `GPG_KEY_NAME` to be set as Secrets

CC: @SantiagoTorres @dlorenc We should also have some information about how to retrieve the public key and steps that can be used to validate the packages.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,10 +26,28 @@ jobs:
         with:
           version: v3.7.1
 
+      - name: Install sigstore Helm plugin
+        run: |
+          helm plugin install https://github.com/sigstore/helm-sigstore
+      - name: Install GPG Keys
+        run: |
+          cat <(echo -e "${{ secrets.GPG_PRIVATE_KEY }}") | gpg --import --batch
+          gpg --export > /home/runner/.gnupg/pubring.gpg
+          gpg --export-secret-keys > /home/runner/.gnupg/secring.gpg
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@c25b74a986eb925b398320414b576227f375f946 # pin@v1.2.1
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_SIGN: "true"
+          CR_KEY: "${{ secrets.GPG_KEY_NAME }}"
+          CR_KEYRING: "/home/runner/.gnupg/secring.gpg"
+
+      - name: Upload Helm Charts to Rekor
+        run: |
+          for chart in `find .cr-release-packages -name '*.tgz' -print`; do
+            helm sigstore upload --keyring=/home/runner/.gnupg/secring.gpg ${chart}
+          done
 
       - name: Login to GitHub Container Registry
         run: |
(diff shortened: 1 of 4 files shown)


cpanato (approved):
I don't know why I thought that we can use the keys that we generate in cosign instead of gpg keys.

lgtm

dlorenc:
@cpanato not yet - they're not really supported by helm fully. :(

.github/workflows/release.yml
```diff
@@ -26,10 +26,28 @@ jobs:
...
+          cat <(echo -e "${{ secrets.GPG_PRIVATE_KEY }}") | gpg --import --batch
+          gpg --export > /home/runner/.gnupg/pubring.gpg
+          gpg --export-secret-keys > /home/runner/.gnupg/secring.gpg
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@c25b74a986eb925b398320414b576227f375f946 # pin@v1.2.1
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_SIGN: "true"
+          CR_KEY: "${{ secrets.GPG_KEY_NAME }}"
```
SantiagoTorres, line 43: Just to double check, is GPG_KEY_NAME supposed to be a uid from the key, or is it supposed to be the fingerprint?  (just to make sure I fill it properly)

sabre1041: @SantiagoTorres it is the `uid` (can use the id or I typically just include the email address)

.github/workflows/release.yml
```diff
@@ -26,10 +26,28 @@ jobs:
...
           version: v3.7.1
 
+      - name: Install sigstore Helm plugin
+        run: |
+          helm plugin install https://github.com/sigstore/helm-sigstore
+      - name: Install GPG Keys
+        run: |
+          cat <(echo -e "${{ secrets.GPG_PRIVATE_KEY }}") | gpg --import --batch
+          gpg --export > /home/runner/.gnupg/pubring.gpg
+          gpg --export-secret-keys > /home/runner/.gnupg/secring.gpg
```
SantiagoTorres, line 36: Not sure what these two lines are for? the import should do enough as long as you're in the gpg user. Otherwise I think we could use --homedir as a parameter to --import --batch and do away with these two lines.

sabre1041: [Helm documentation](https://helm.sh/docs/topics/provenance/) calls out these steps

SantiagoTorres: Ahh, so it's to force a v1 format. Gotcha!

SantiagoTorres (approved):

sabre1041:
@cpanato Would you be able to review and merge this in? Once integrated, i'll bump both charts so that we can test out the new workflow

sabre1041:
@SantiagoTorres do you have details on how users can import the public key to perform verification on their side?

SantiagoTorres:
Sure. I updated the secrets BTW. Where do you think we should put the information on how to fetch/use the pubkey?

sabre1041:
@SantiagoTorres Since the public key might change per release, we would probably want to have the specific public key included in each release (chart releaser creates a tagged release for every published chart). we could include details in the release notes. And then a note in the README that calls out how to verify signed charts pointing to the release notes.

Thoughts?

SantiagoTorres:
This LGTM! Shall I click the green button?

sabre1041:
@SantiagoTorres go for it! whats the worst that can happen :)

PR #49, opened 2021-12-21
docs(cosigned): enhance doc with k8s:// scheme

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

#### Summary

enhance doc with k8s:// scheme

#### Ticket Link

Fixes

#### Release Note

```release-note
docs(cosigned): enhance doc with k8s:// scheme
```

### DIFF
--- a/charts/cosigned/README.md
+++ b/charts/cosigned/README.md
@@ -7,18 +7,13 @@
 ### Deploy `cosigned` Helm Chart
 
 
+Let's create a secret to validate the signatures, once we define `k8s://` scheme for the command, cosign will generate public/private key pairs, and store them in a Secret without touching a disk:
 ```shell
 export COSIGN_PASSWORD=<my_cosign_password>
-cosign generate-key-pair
-```
 
-The previous command generates two key files `cosign.key` and `cosign.pub`. Next, create a secret to validate the signatures:
-
-```shell
 kubectl create namespace cosign-system
 
-kubectl create secret generic mysecret -n \
-cosign-system --from-file=cosign.pub=./cosign.pub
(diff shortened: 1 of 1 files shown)


charts/cosigned/README.md
```diff
@@ -7,18 +7,13 @@
...
-```
-
-The previous command generates two key files `cosign.key` and `cosign.pub`. Next, create a secret to validate the signatures:
 
-```shell
 kubectl create namespace cosign-system
 
-kubectl create secret generic mysecret -n \
-cosign-system --from-file=cosign.pub=./cosign.pub
+cosign generate-key-pair k8s://cosign-system/mysecret
```
cpanato, line 16: in the secret, we just need to pub key afaik, not sure if we want to have the private key together.
thoughts @hectorj2f ?

hectorj2f: Yes, I won't suggest adding both for security reasons. That is the intention of only adding the public key.

hectorj2f: It could also mislead users which simply copy and paste these instructions, and come back asking why we need the private key.

developer-guy: In the current design in cosign, we store both passwords, private and public keys in Kubernetes Secret btw.

hectorj2f: > In the current design in cosign, we store both passwords, private and public keys in Kubernetes Secret btw.

Yes. But cosigned doesn't need a private key to work.

dlorenc: There was actually some confusion around this in slack a few weeks ago - it's best to keep the private key out of here!

PR #51, opened 2021-12-28
rekor: bump to release v0.4.0

cpanato:
#### Summary
- rekor: bump to release v0.4.0

#### Ticket Link
n/a

#### Release Note

```release-note
rekor: bump to release v0.4.0
```

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,8 +4,8 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.9
-appVersion: 0.3.0
+version: 0.2.10
+appVersion: 0.4.0
 
 keywords:
   - security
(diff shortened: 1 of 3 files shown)


charts/rekor/Chart.yaml
```diff
@@ -4,8 +4,8 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.9
-appVersion: 0.3.0
+version: 0.3.0
```
sabre1041, line 7: Is there a need to bump to the next minor chart version given that the changes here are associated to the images within the chart?

cpanato: this is a similar approach we did in helm/charts in the past and in other charts, when there is a minor version release of the app/service we bump the minor in the chart as well because that is not a patch fix.
however, I'm up to you if you don't want to do that.

sabre1041: I think moving forward, we should adopt only increasing the chart minor version for significant change. If we could go with chart version of `0.2.10`, thats what I'd recommend

cpanato: updated. ptal
thanks for the review and feedback

sabre1041 (changes requested):
LGTM

sabre1041 (approved):

PR #59, opened 2022-02-11
rekor: Make some components depends on "enabled"

ohkinozomu:
#### Summary

Some components are created regardless of the value of `enabled`.
This PR makes them not created when `enabled` is false.

Since the MySQL secret is referenced by trillian, I leave it to be created regardless of the "enabled" setting.

This change may remove some components already created, so the patch version increment may be inappropriate. Please advise.

#### Release Note

```release-note
Make some components depends on "enabled"
```


ohkinozomu:
@sabre1041 
OK, feel free to close this PR when it becomes outdated.
I look forward to the upcoming changes.

sabre1041:
@ohkinozomu should now be included in #62

ohkinozomu:
@sabre1041 
Thank you for the great job!
However, let me point out here that "enabled" is not in effect in some new charts.

## ctlog
`createtree.enabled`
`createctconfig.enabled`

## fulcio
`createcerts.enabled`

## rekor
`server.enabled`

## trillian
`logServer.enabled`
`logSigner.enabled`
`mysql.enabled` doesn't affect mysql deployment.

PR #62, opened 2022-02-28
Scaffolding sigstore architecture

sabre1041:
This integration will provide a method to sigstore ecosystem:

Capabilities include:

* Creation of new charts
    * ctlog
    * fulcio
* Split out of trillian from rekor as a new standalone chart
* Umbrella chart called `scaffold` that enables standing up the entire sigstore architecture
* Updates of chart runner


charts/ctlog/values.yaml
```diff
@@ -0,0 +1,59 @@
+namespace:
+  create: false
+  name: ctlog-system
+server:
+  replicaCount: 1
+  image:
+    registry: ghcr.io
+    repository: vaikas/ct_server
+    pullPolicy: IfNotPresent
```
vaikas, line 9: I think this might cause hilarity on Kind installations, so wonder if we should just remove this?

charts/ctlog/values.yaml
```diff
@@ -0,0 +1,59 @@
+namespace:
+  create: false
+  name: ctlog-system
+server:
+  replicaCount: 1
+  image:
+    registry: ghcr.io
+    repository: vaikas/ct_server
```
vaikas, line 8: We might add a comment / whatever somewhere putting a pointer in readme or something that points to the repo, or explains why we're pulling these from my repo and say it's being upstreamed as much as we can, just takes a bit of time.

vaikas (commented):
Thank you so much for pulling all these together!!!

charts/ctlog/values.yaml
sabre1041: README could work

cpanato: should we move those to the sigstore or another place and create some automation to publish the images? instead of using a personal repository/registry

cpanato:
+1 on `sigstore/scaffolding`

vaikas:
Today during the sigstore community call, I asked about this. I created this to see if we can get that repo:
https://github.com/sigstore/tac/issues/5

sabre1041:
@cpanato @vaikas @strongjz @dlorenc This should be good to go!

charts/scaffold/README.md
```diff
@@ -0,0 +1,77 @@
+# scaffold
+
+Umbrella chart for deploying the components of the [sigstore](https://www.sigstore.dev/) architecture.
+
+The following components are also included as either direct components or through chart dependencies:
+
+* ctlog
```
hectorj2f, line 7: Perhaps it is worthy to add a link to https://github.com/google/certificate-transparency-go here.

hectorj2f (approved):

charts/scaffold/README.md
sabre1041: Added

cpanato (approved):
thanks!!

PR #69, opened 2022-03-18
fix: ingress service name not found

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

#### Summary

Ingress service name `scaffold-rekor-server` does not match the service name of the rekor server which was `rekor`. Another option would be to change the helpers.tpl so the service name is `rekor.fullname` instead of `rekor.server.fullname`.

Screenshot 2022-03-18 at 14 52 55

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (approved):
thanks

PR #79, opened 2022-03-28
Fix rekor chart lint error

k4leung4:
Do not trim trailing whitespace for conditional
server.logging.production, as it causes the value to be appended to the
previous line instead of the next.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.15
+version: 0.2.16
 appVersion: 0.5.0
 
 keywords:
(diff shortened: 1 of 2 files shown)


charts/rekor/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.2.15
+version: 0.2.16
```
sabre1041, line 7: @k4leung4 Can the version be bumped on the scaffold chart with this change?

sabre1041 (changes requested):

charts/rekor/Chart.yaml
k4leung4: done, thanks for pointing that out.

sabre1041:
@k4leung4 sigh. alright. looks like we have to release rekor first and then bump scaffold. If you want to revert the change, i'll push it through

k4leung4:
> @k4leung4 sigh. alright. looks like we have to release rekor first and then bump scaffold. If you want to revert the change, i'll push it through

thanks, just figured that out.
reverted.

sabre1041:
@k4leung4 you might want to just rebase into 1 commit so chart tester will play along

sabre1041 (approved):

PR #82, opened 2022-03-29
Allow specifying multiple hosts for service TLS

k4leung4:

#### Summary

    Add separate property tlsEnabled to control population of tls fields.
    Change existing tls type from object to array to mirror k8s ingress
    schema. allowing multiple tls/hosts to be specified.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/rekor/templates/server/ingress.yaml
+++ b/charts/rekor/templates/server/ingress.yaml
@@ -10,18 +10,24 @@ metadata:
 {{ toYaml .Values.server.ingress.annotations | indent 4 }}
 spec:
   rules:
-    - host: {{ required "An Ingress hostname is required" .Values.server.ingress.hostname }}
+    {{- range .Values.server.ingress.hosts }}
+    - host: {{ required "An Ingress hostname is required" .host | quote }}
       http:
         paths:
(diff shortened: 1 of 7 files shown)


charts/rekor/templates/server/ingress.yaml
```diff
@@ -10,18 +10,24 @@ metadata:
 {{ toYaml .Values.server.ingress.annotations | indent 4 }}
 spec:
   rules:
-    - host: {{ required "An Ingress hostname is required" .Values.server.ingress.hostname }}
+    {{- range .Values.server.ingress.hosts }}
```
cpanato, line 13: it is missing the ingress.hosts in the values.yaml and also will need to add it in `ci/ci-values.yaml` (maybe need to remove the exiting hostname)

cpanato (changes requested):
there are some missing/updated values that need to be added

charts/rekor/templates/server/ingress.yaml
k4leung4: added ingress.hosts, thanks for pointing that out.
updated ci/ci-values.yaml

ptal

PR #84, opened 2022-03-29
Move prometheus annotations from deployment to pod for Rekor.

k4leung4:
Prometheus scrap annotations should be on the pod and not the
deployment.

Also clean up some of the extra blank lines.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/rekor/values.yaml
+++ b/charts/rekor/values.yaml
@@ -120,7 +120,7 @@ server:
       existingClaim: ""
       accessModes:
         - ReadWriteOnce
-  deploymentAnnotations:
+  podAnnotations:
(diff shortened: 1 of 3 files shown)


charts/rekor/values.yaml
```diff
@@ -120,7 +120,7 @@ server:
       existingClaim: ""
       accessModes:
         - ReadWriteOnce
-  deploymentAnnotations:
+  podAnnotations:
```
cpanato, line 123: we might add a service monitor in the future in case people we using prometheus operator. i will follow up on that

PR #90, opened 2022-04-01
bump scaffold version for new rekor/fulcio version

k4leung4:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/scaffold/Chart.yaml
+++ b/charts/scaffold/Chart.yaml
@@ -4,7 +4,7 @@ description: Scaffolding the components of the sigstore architecture
 
 type: application
 
-version: 0.1.3
+version: 0.1.4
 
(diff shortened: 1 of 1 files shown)


cpanato (changes requested):
you need to bump the versions for each chart in a different PR, otherwise when the release happens will break it because we don't know if the fulcio/rekor will run first and scafold will run in the end

can you split it? but the question is why just bump fulcio/rekor with no changes

k4leung4:
i was thinking i need bump the version when changing dependency version, I guess that is not needed?

cpanato:
i think it is all using the latest charts, looks like this PR is not needed

cpanato (approved):

PR #108, opened 2022-04-05
default mysql hostname to localhost when using gcp mysql

k4leung4:

#### Summary

default mysql hostname to localhost when using gcp mysql
Fix bad cloud sql proxy image tag. missing patch .0, for 1.29.0

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/trillian/templates/_helpers.tpl
+++ b/charts/trillian/templates/_helpers.tpl
@@ -27,7 +27,11 @@ If release name contains chart name it will be used as a full name.
 Return the hostname for mysql
 */}}
 {{- define "mysql.hostname" -}}
+{{- if (and (not .Values.mysql.enabled) .Values.mysql.gcp.enabled) -}}
+{{- default "localhost" .Values.mysql.hostname }}
+{{- else -}}
 {{- default (include "trillian.mysql.fullname" .) .Values.mysql.hostname }}
+{{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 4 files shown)


charts/trillian/templates/_helpers.tpl
```diff
@@ -27,8 +27,12 @@ If release name contains chart name it will be used as a full name.
 Return the hostname for mysql
 */}}
 {{- define "mysql.hostname" -}}
+{{- if (and (not .Values.mysql.enabled) .Values.mysql.gcp.enabled) -}}
+localhost
```
sabre1041, line 31: Can the hostname be provided through the `mysql.hostname` value. Is there a reason to make special considerations for GCP? Will it always be `localhost` when deployed to GCP? If not, its not overridable

k4leung4: thanks for the feedback.
fixed the logic to default to "localhost" if mysql.hostname is not specified, so the user can always still override it as necessary.

when mysql.gcp.enabled=true, the google cloud sql proxy sidecar is added to run along trillian log server and signer. and to use the proxy sidecar, one would use `localhost`. so it feels like with mysql.gcp.enabled=true, we are expecting users to use the sidecar approach.

this is just a nice to have at the end of the day, and im fine if this doesnt go in.  the aim was to make the user experience a bit better.

sabre1041: I'd recommend holding off on this one as it becomes somewhat sticky to a particular cloud/implementation. @cpanato  thoughts?

k4leung4: ack, thanks for taking time to look over it.

k4leung4:
closing for now.

Issue #109, opened 2022-04-07
Use string field for Fulcio config content


nsmith5:
**Description**

Currently Fulcio chart uses an object for the Fulcio config content (`.Values.config.content`). This means that values are merged with default instead of overriding defaults when applied. The result is that its impossible to remove the default OIDC issuer of `https://kubernetes.default.svc` using the Fulcio chart.

We should use a string for this field so that changes completely override the config. Eg

```diff
config:
+ content: |-
- content:
    { "foo": "bar" }
```

PR #115, opened 2022-04-08
move default into template helper to allow overwrite.

k4leung4:

#### Summary

Due to the nature of helm handling of maps, it is not possible to
overwrite keys in the base value.yaml of the chart. Keys from value.yaml
included with the chart will always be present.

If it was the main chart, it would be possible to null existing keys,
but due to https://github.com/helm/helm/issues/9136, it does not work
with sub-charts at the moment.

So to allow downstream users to overwrite the config contents, the
default is now set in the template helper, when an empty object is
specified.

@nsmith5 what do you think about this solution? 

#### Ticket Link

Fixes https://github.com/sigstore/helm-charts/issues/109

this addresses the issue https://github.com/sigstore/helm-charts/issues/109 without changing config.content to string type.

#### Release Note

```release-note

```

### DIFF
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -4,7 +4,7 @@ description: Fulcio
 
 type: application
 
-version: 0.2.7
+version: 0.2.8
 
 keywords:
   - security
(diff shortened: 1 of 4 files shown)


nsmith5 (approved):

nsmith5:
I like it!

k4leung4:
@cpanato please take a look when you get a chance, thank you.

cpanato (approved):
if @nsmith5 likes I like too 😃

PR #131, opened 2022-04-15
Add ingress for ctlog helm.

k4leung4:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/ctlog/templates/_helpers.tpl
+++ b/charts/ctlog/templates/_helpers.tpl
@@ -188,3 +188,46 @@ Create Container Ports based on Service Ports
   protocol: {{ default "TCP" .protocol }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "ingress.apiVersion" -}}
+{{- if .Values.server.ingress.apiVersion -}}
+{{- .Values.server.ingress.apiVersion -}}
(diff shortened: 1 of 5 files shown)


priyawadhwa (approved):

charts/ctlog/templates/_helpers.tpl
```diff
@@ -188,3 +188,46 @@ Create Container Ports based on Service Ports
...
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "ingress.apiVersion" -}}
+{{- if .Values.server.ingress.apiVersion -}}
+{{- .Values.server.ingress.apiVersion -}}
+{{- else if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version -}}
```
hectorj2f, line 198: do we need to support kubernetes version < 1.14 ?

k4leung4: I personally dont think we need to support < 1.14, though this logic is copied from rekor/fulcio, and prefer to keep it consistent with the support.
If we deem we want to remove support for <1.14, I'm happy to remove them from all three charts in a future pr.

hectorj2f: Alright

hectorj2f (approved):

PR #134, opened 2022-04-18
bump rekor helm chart for 0.6.0

k4leung4:
bump createtree to 0.2.6, from 0.2.1
bump redis to 6.2.6-alpine3.15, from 5.0.10
bump curl to 7.82.0, from 7.81.0

Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

PR #135, opened 2022-04-18
Update rekor chart to v0.6

priyawadhwa:

Adds the sharding configmap as a helm template, which is a required configmap for rekor >= v0.6

```release-note
Update rekor chart to v0.6
```

### DIFF
--- a/charts/rekor/templates/server/configmap-sharding.yaml
+++ b/charts/rekor/templates/server/configmap-sharding.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "rekor.sharding-config" . }}
+{{ include "rekor.namespace" . | indent 2 }}
+  labels:
+    {{- include "rekor.server.labels" . | nindent 4 }}
+data:
+  {{ .Values.server.sharding.filename }}: |
+    {{ .Values.server.sharding.contents }}
--- a/charts/rekor/templates/server/deployment.yaml
+++ b/charts/rekor/templates/server/deployment.yaml
@@ -64,6 +64,8 @@ spec:
           volumeMounts:
             - name: storage
               mountPath: {{ template "rekor.server.fileAttestationStorage.path" . }}
+            - name: {{ template "rekor.sharding-config" . }}
+              mountPath: {{ .Values.server.sharding.mountPath }}
 {{- end -}}
 {{- if .Values.server.livenessProbe }}
           livenessProbe:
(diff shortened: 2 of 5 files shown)


charts/rekor/templates/server/configmap-sharding.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.server.sharding.name }}
```
k4leung4, line 4: for consistency, can you change to generate the name similar to existing configmap, with template "rekor.config.

https://github.com/sigstore/helm-charts/blob/main/charts/rekor/templates/server/configmap.yaml#L4
https://github.com/sigstore/helm-charts/blob/main/charts/rekor/templates/_helpers.tpl#L410

charts/rekor/templates/server/deployment.yaml
```diff
@@ -57,13 +57,16 @@ spec:
                   name: {{ template "rekor.config" . }}
                   key: {{ .Values.server.config.key }}
           args:
+          - "--trillian_log_server.sharding_config={{ .Values.server.sharding.mountPath }}/{{ .Values.server.sharding.filename }}"
```
k4leung4, line 60: can you move this into the template helper?

https://github.com/sigstore/helm-charts/blob/main/charts/rekor/templates/_helpers.tpl#L301

charts/rekor/values.yaml
```diff
@@ -93,6 +93,12 @@ server:
     httpGet:
       port: 3000
       path: /ping
+  sharding:
+    name: sharding-config
```
k4leung4, line 97: do we want to add a bit to allow enabling/disabling sharding?

k4leung4 (changes requested):

charts/rekor/values.yaml
priyawadhwa: The sharding configmap is required to exist, even if it's empty. There isn't really an option to "disable" it.

charts/rekor/templates/server/configmap-sharding.yaml
priyawadhwa: 👍🏽

charts/rekor/values.yaml
k4leung4: ack, thanks for the clarification.

k4leung4:
it failed with `Error: unknown flag: --trillian_log_server.sharding_config`

you will want to update the image to 0.6.0 image in your change to fix this.
override the digest in values.yaml, and update the image in Charts.yaml

priyawadhwa:
thanks, i pretty much just copied over your changes from #134!

k4leung4 (approved):
thank you for updating rekor chart to 0.6.0

PR #141, opened 2022-04-19
Add support for grpc in fulcio.

k4leung4:

#### Summary

This is a breaking change, as it adds another level for ingress, http
and grpc.

I am open to making it backwards compatible by keeping the http settings
at the same level, and have all of the grpc settings one level deeper.
The downside to this is it makes it messier having a mix of settings.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -4,7 +4,7 @@ description: Fulcio
 
 type: application
 
-version: 0.2.11
+version: 0.3.0
 
 keywords:
   - security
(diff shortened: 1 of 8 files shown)


charts/fulcio/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Fulcio
 
 type: application
 
-version: 0.2.11
+version: 0.2.12
```
cpanato, line 7: if is a breaking change I would bump it to 0.3.0 and add a readme doc to alert that, and point what things will be need to re-generated

k4leung4: good advise. 
done.

priyawadhwa (approved):

hectorj2f (approved):

cpanato (approved):

PR #142, opened 2022-04-20
Add option to store CT log config in a secret

priyawadhwa:
Right now the CT log config has to be stored in a configmap. This PR also adds the option to store it in a Secret, leaving the ConfigMap as the default.

This will make it possible to store the config externally & use something like external-secrets to access the config. The config also stores the password for the ct log private key decryption, so it might make more sense in a Secret.

### DIFF
--- a/charts/ctlog/Chart.yaml
+++ b/charts/ctlog/Chart.yaml
@@ -4,7 +4,7 @@ description: Certificate Log
 
 type: application
 
-version: 0.2.11
+version: 0.2.12
 
 keywords:
   - security
--- a/charts/ctlog/templates/ctlog-deployment.yaml
+++ b/charts/ctlog/templates/ctlog-deployment.yaml
@@ -55,8 +55,16 @@ spec:
               - key: rootca
                 path: roots.pem
         - name: config
+          {{- if .Values.server.config.secret.enabled }}
+          secret:
+            secretName: {{ .Values.server.config.secret.name }}
+            items:
+              - key: config
+                path: ct_server.cfg
+          {{ else }}
           configMap:
             name: {{ template "ctlog.config" . }}
             items:
               - key: config
                 path: ct_server.cfg
+          {{- end }}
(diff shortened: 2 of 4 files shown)


k4leung4 (changes requested):
looks good.
if the configmap isnt used, please make the createctconfig job/serviceaccount conditional on the secret enabled bit, as well as the cm-operator-role[binding] used to create the cm.

charts/ctlog/templates/secret-operator-rolebinding.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if not .Values.server.config.secret.enabled }}
```
k4leung4, line 1: should we do this for secret-operator-role.yaml as well?

priyawadhwa: yup sg, also added it to `ctlog-configmap.yaml`

charts/ctlog/templates/createctconfig-job.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if not .Values.server.config.secret.enabled }}
```
k4leung4, line 1: really sorry, looks like we have to keep the createctconfig and configmap.

createctconfig also creates the secret with the pub key, and store the treeID in the configmap.

even though we dont need the config data, there is no way to keep the others without this.
so i propose keeping everything, like you originally had in your PR. really sorry.
and filing an issue to make that part of the createctconfig configurable by flag to not generate the config.

priyawadhwa: no worries, it's easy to switch back! created an issue (https://github.com/sigstore/scaffolding/issues/129) but specified we could add a flag to createctconfig to store the config in a secret, which would go well with this option being added to helm

k4leung4 (approved):

PR #167, opened 2022-05-14
Add in a helm chart for the sigstore prober

priyawadhwa:

 **Description of the change**

 **Additional Information**

 

**Checklist**

- [x ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). Where applicable, update and bump the versions in any associated umbrella chart
- [ x] Variables are documented in the `values.yaml` and added to the README.md. THe [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-doc --dry-run` to preview the content
- [ x] JSON Schema generated
- [ x] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command

### DIFF
--- a/charts/sigstore-prober/Chart.yaml
+++ b/charts/sigstore-prober/Chart.yaml
@@ -1,0 +1,22 @@
+apiVersion: v2
+name: sigstore-prober
+description: Sigstore API Endpoint Prober
+
+type: application
+
+version: 0.0.1
+
+keywords:
+  - security
+  - prober
+
+home: https://sigstore.dev/
+
+maintainers:
+  - name: The Sigstore Authors
+
+annotations:
+  artifacthub.io/license: Apache-2.0
+  artifacthub.io/images: |
+    - name: sigstore-prober
+      image: gcr.io/priya-chainguard/prober-37ee855f344691c13c5769b1675cc9e0@sha256:a601f7de3c572bb65b2cebd431ee8fc0cbd409b2edb5836718f09973400538c8
--- a/charts/sigstore-prober/Chart.yaml
+++ b/charts/sigstore-prober/Chart.yaml
@@ -5,6 +5,8 @@ description: Sigstore API Endpoint Prober
 type: application
 
 version: 0.0.1
+appVersion: 0.0.1
+
 
 keywords:
   - security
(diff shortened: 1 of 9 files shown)


charts/sigstore-prober/values.yaml
```diff
@@ -0,0 +1,17 @@
+namespace:
+  create: true
+  name: sigstore-prober
+spec:
+  replicaCount: 1
+  serviceAccountName: default
```
hectorj2f, line 6: Question: aren't you missing the creation of the serviceAccount ?

priyawadhwa: Since the service account requires workload identity enabled, I thought it might make more sense for folks to create it separately. But I can add it in if that would be useful!

hectorj2f: It was just to be consistent with the rest of the charts, as we also defined the SA. https://github.com/sigstore/helm-charts/blob/main/charts/ctlog/templates/ctlog-serviceaccount.yaml

hectorj2f (approved):

charts/sigstore-prober/templates/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if .Values.namespace.create }}
```
cpanato, line 1: i don't think we need to set this because we can use the helm cli to create or point to the namespace we want to use.

but also i don't have a strong opinion on this

charts/sigstore-prober/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v2
+name: sigstore-prober
+description: Sigstore API Endpoint Prober
+
+type: application
+
+version: 0.0.1
```
cpanato, line 7: need to add the `appVersion `

charts/sigstore-prober/templates/namespace.yaml
sabre1041: This would follow the pattern implemented by other charts in the repository. This parameter was designed to support the scaffold chart

cpanato: ahh, i forgot those 🤦

charts/sigstore-prober/values.yaml
priyawadhwa: Thanks, added it in!

charts/sigstore-prober/Chart.yaml
priyawadhwa: done :)

hectorj2f (approved):
Lgtm

PR #194, opened 2022-06-06
have ctlog deployment rollout with configmap change.

k4leung4:

 **Description of the change**

to ensure deployment will always use the latest configmap.

following https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

@nsmith5 PTAL

 **Additional Information**

 

**Checklist**

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). Where applicable, update and bump the versions in any associated umbrella chart
- [ ] Variables are documented in the `values.yaml` and added to the README.md. THe [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-docs --dry-run` to preview the content
- [ ] JSON Schema generated
- [x] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command

### DIFF
--- a/charts/ctlog/templates/ctlog-deployment.yaml
+++ b/charts/ctlog/templates/ctlog-deployment.yaml
@@ -12,10 +12,11 @@ spec:
       {{- include "ctlog.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-    {{- if .Values.server.podAnnotations }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/ctlog-configmap.yaml") . | sha256sum }}
+        {{- if .Values.server.podAnnotations }}
         {{- toYaml .Values.server.podAnnotations | nindent 8 }}
-    {{- end }}
+        {{- end }}
       labels:
         {{- include "ctlog.selectorLabels" . | nindent 8 }}
     spec:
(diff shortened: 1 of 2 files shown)


cpanato (dismissed):

charts/ctlog/templates/ctlog-deployment.yaml
```diff
@@ -14,6 +14,7 @@ spec:
     metadata:
     {{- if .Values.server.podAnnotations }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/ctlog-configmap.yaml") . | sha256sum }}
         {{- toYaml .Values.server.podAnnotations | nindent 8 }}
```
nsmith5, line 18: I think the if block needs to move closer to the where the pod annotations are being templated now so that a lack of pod annotations does kill the checksum annotation: 

```suggestion
        {{- if .Values.server.podAnnotations }}
        {{- toYaml .Values.server.podAnnotations | nindent 8 }}
        {{- end }}
```

nsmith5 (changes requested):
Checksum logic looks good, just need to move that if block I think

charts/ctlog/templates/ctlog-deployment.yaml
k4leung4: thanks for catching that.

fixed.

nsmith5 (approved):
lgtm!

PR #264, opened 2022-08-15
Add rekor-createtree helm chart for creating a new tree for rekor

priyawadhwa:
This will be used for sharding. Initially, I thought we could just force the `createtree` Job through the rekor Chart.
But, it looks like that will automatically restart the rekor server, which might lead to a race between updating the Rekor configmap with the new tree ID and starting up the server.

With a separate Chart, we can create a new tree, update the ConfigMap, but still  control when we actually restart the rekor server to pick up that change.

### DIFF
--- a/charts/rekor-createtree/.helmignore
+++ b/charts/rekor-createtree/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/rekor-createtree/Chart.yaml
+++ b/charts/rekor-createtree/Chart.yaml
@@ -1,0 +1,21 @@
+apiVersion: v2
+name: rekor-createtree
+description: Part of the sigstore project, rekor-createtree is used to create new Trillian trees for Rekor.
+
+type: application
+
+version: 0.0.1
+
+keywords:
+  - security
+  - transparency logs
+
+home: https://sigstore.dev/
+maintainers:
+  - name: The Sigstore Authors
+
+annotations:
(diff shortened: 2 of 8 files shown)


k4leung4:
I believe the rekor-server is restarting as it watches the configmap. any time when the configmap changes, rekor-server will restart to read in the new configmap changes.

see #195 

this was added as the configmap changes would be picked up when the pod restarted, which may be a surprise.

if we want full control over which configmap to use and when, should we use a different name for the configmap?

> This will be used for sharding. Initially, I thought we could just force the `createtree` Job through the rekor Chart. But, it looks like that will automatically restart the rekor server, which might lead to a race between updating the Rekor configmap with the new tree ID and starting up the server.
> 
> With a separate Chart, we can create a new tree, update the ConfigMap, but still control when we actually restart the rekor server to pick up that change.
> 
> Signed-off-by: Priya Wadhwa [priya@chainguard.dev](mailto:priya@chainguard.dev)

priyawadhwa:
> if we want full control over which configmap to use and when, should we use a different name for the configmap?

yes that could work! alternatively, we could have the rekor deployment rollout on sharding config map changes as well. then the flow could be something like this:

1. scale down rekor to 1 pod
2. in one PR, we can update the sharding configmap with the "old" tree ID & also set `createtree.force=true`. then, once the new tree ID is created the rekor server will automatically be restarted.
3. scale rekor back to 3 pods

priyawadhwa:
Did some manual testing I think that will work, will close this.

k4leung4:
> > if we want full control over which configmap to use and when, should we use a different name for the configmap?
> 
> yes that could work! alternatively, we could have the rekor deployment rollout on sharding config map changes as well. then the flow could be something like this:
> 
> 1. scale down rekor to 1 pod
> 2. in one PR, we can update the sharding configmap with the "old" tree ID & also set `createtree.force=true`. then, once the new tree ID is created the rekor server will automatically be restarted.
> 3. scale rekor back to 3 pods

that sounds like it should work.

btw, im not against creating another helm chart, i just want to be clear that if we point to the same configmap, whether it is the same or different helm chart, rekor server would auto-rollout. though it would be nice to avoid an additional chart.

k4leung4:
thanks

priyawadhwa:
> though it would be nice to avoid an additional chart.

yah i feel the same way! i don't think this will add much now that i know the deployment restarts if the configmap changes (initially i thought that was just something helm did since they were coupled in the same chart)

PR #279, opened 2022-08-31
add gce lb files for rekor

bobcallaway:
Signed-off-by: Bob Callaway <bcallaway@google.com>

### DIFF
--- a/charts/rekor/README.md
+++ b/charts/rekor/README.md
@@ -1,34 +1,26 @@
 # rekor
 
-[Rekor](https://docs.sigstore.dev/rekor/overview/) provides a restful API based server for validation and a transparency log for storage.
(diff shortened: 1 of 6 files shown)


cpanato:
i think we should set here https://github.com/sigstore/helm-charts/blob/main/charts/rekor/ci/ci-values.yaml the className to be nginx, the kind cluster does not have the crds for the gcp ingress

bobcallaway:
> i think we should set here https://github.com/sigstore/helm-charts/blob/main/charts/rekor/ci/ci-values.yaml the className to be nginx, the kind cluster does not have the crds for the gcp ingress

should we do that or install the crds (assuming we can find them)?

charts/rekor/README.md
```diff
@@ -162,25 +159,3 @@ The following table lists the configurable parameters of the Rekor chart and the
 | trillian.namespace.create | bool | `true` |  |
 | trillian.namespace.name | string | `"trillian-system"` |  |
 
-----------------------------------------------
```
k4leung4, line 165: should we preserve this section?

PR #281, opened 2022-09-02
Adjustments to account for home directory and adjusting probe for Policy Controller

### LINKED ISSUE #263: policy-controller 0.2.1 read-only file system errors
**Description**
Hi,
I've been trialling policy-controller on a small k8s cluster on digitalocean, trying to validate an image with a keyless signature, built and signed on github and hosted in a private registry (digitalocean registry).

Basing it on the instructions from [the chart's readme](https://github.com/sigstore/helm-charts/tree/main/charts/policy-controller) and the documentation, I installed the controller with:
```
helm upgrade --install --create-namespace -n cosign-system --version 0.2.1 policy-controller sigstore/policy-controller --set policywebhook.resources.requests.cpu="50m" --set webhook.resources.requests.cpu="50m"
```
(pin version instead of devel, skip the secret that as far as I understand it's optional fallback behaviour, change limits b/c my worker nodes are really small).

Once the pods are available I add a policy like so:

```
apiVersion: policy.sigstore.dev/v1beta1
kind: ClusterImagePolicy
metadata:
  name: image-policy
spec:
  images:
  - glob: "**"
  authorities:
    - keyless:
        identities:
          - issuer: https://token.actions.githubusercontent.com
            subjectRegExp: https://github.com/MYORG/\S+/.github/workflows/\S+@refs/heads/main
```
and enable on a namespace adding label as per docs.
I then test by applying a deployment which pod image is keyless-signed via GH workflow (and stored in private registry). I would expect a pass, but instead the error I get back is:

```
Error from server (BadRequest): error when creating "test-deploy.yml": admission webhook "policy.sigstore.dev" denied the request: validation failed: failed policy: image-policy: spec.template.spec.containers[0].image
registry.digitalocean.com/guanciale/servatrice@sha256:REDACTED signature keyless validation failed for authority authority-0 for registry.digitalocean.com/MYORG/MYIMAGE@sha256:REDACTED: no matching signatures:
creating cached local store: mkdir /.sigstore: read-only file system
 creating cached local store: mkdir /.sigstore: read-only file system
```
I would have expected a reasonably plain setup with the latest version to only fail with errors that suggest my config changes might be at fault - but I can't see how this one is related.

I'm not sure what I might be doing wrong here or where to look for, any suggestion welcome. 

**Version**

policy-controller chart `0.2.1`

sabre1041:
Corrections to handle permissions within Policy Controller

* Add `emptyDir` for ephemeral storage
* Set `HOME` environment variable so that all containers respect the `emptyDir` volume for ephemeral storage
* Correction to probes

Resolves #263

### DIFF
--- a/charts/policy-controller/Chart.yaml
+++ b/charts/policy-controller/Chart.yaml
@@ -8,7 +8,7 @@ sources:
 type: application
 
 name: policy-controller
-version: 0.2.1
+version: 0.2.2
 appVersion: 0.2.1
 
 maintainers:
--- a/charts/policy-controller/templates/policy-webhook/deployment_policy_webhook.yaml
+++ b/charts/policy-controller/templates/policy-webhook/deployment_policy_webhook.yaml
@@ -106,7 +106,7 @@ spec:
           # Failing to provide a writable $HOME can cause TUF client initialization to panic
           - mountPath: /home/nonroot
             name: writable-home-dir
-        readinessProbe: &probe
+        readinessProbe:
           failureThreshold: 6
           initialDelaySeconds: 30
           periodSeconds: 1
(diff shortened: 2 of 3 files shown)


hectorj2f (approved):

hectorj2f:
Thanks @sabre1041

lukehinds (approved):

PR #287, opened 2022-09-09
Expose write-prober flag through sigstore-prober helm-chart.

var-sdk:
Also bumps the prober version to latest.

Signed-off-by: Simon Kent <simek@google.com>


priyawadhwa (approved):
LGTM!

Issue #298, opened 2022-09-16
Shouldn't the Chart.lock files be committed to pin the dependencies?


marcofranssen:
**Question**

Usually it is a good practice to commit lock files to ensure we pin to exactly the same version of a release. E.g. a tag could be replaced and therefore silently install another dependency.

I would propose to remove the exclusion from the Chart.lock from the .gitignore and pin all charts with dependencies by running.

```shell
helm dependency build
```

This will download a tgz, which we should add to the .gitignore.

Happy to provide a PR if you agree.

cpanato:
sounds good, go ahead

sabre1041:
@marcofranssen Ive gone back/forth on whether they should be committed or not. Mainly because at the early stages of the project, there were a lot of pieces still in flux.

However, given that there is relative stability at this time, its a good idea to move forward with that plan.

Thanks for contributing this and some of the other areas that you have proposed for the repository!

PR #297, opened 2022-09-16
bump image versions for trillian

marcofranssen:
**Description of the change**

Bumps the rekor and trillian images to most recent versions.

The current version of the rekor and trillian charts have an issue with the 

mysql and redis container both reporting following error.

```shell
exec /usr/local/bin/docker-entrypoint.sh: exec format error
```

Bumping the images to latest versions at least resolves this.

> **Note**: Also included helm-docs template for rekor and trillian to ease updating the README.md. In values.yaml I also fixed some comments so they count as helm-docs to they show up in the README.md

 **Additional Information**

 

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). Where applicable, update and bump the versions in any associated umbrella chart
- [x] Variables are documented in the `values.yaml` and added to the README.md. THe [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-docs --dry-run` to preview the content
- [x] JSON Schema generated (N.A.)
- [x] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command

PR #299, opened 2022-09-16
Dependency pinning via Chart.lock + documentation improvements.

marcofranssen:
**Description of the change**

#298 

 **Additional Information**

 

**Checklist**

- [x] Variables are documented in the `values.yaml` and added to the README.md. THe [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-docs --dry-run` to preview the content
- [x] JSON Schema generated
- [x] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -1,6 +1,8 @@
 apiVersion: v2
 name: rekor
-description: Part of the sigstore project, Rekor is a timestamping server and transparency log for storing signatures, as well as an API based server for validation
+description: |
+  Part of the Sigstore project, Rekor is a timestamping server and transparency
+  log for storing signatures, as well as an API based server for validation.
 
 type: application
 
@@ -14,8 +16,11 @@ keywords:
 home: https://sigstore.dev/
 sources:
   - https://github.com/sigstore/rekor
+  - https://github.com/sigstore/helm-charts/tree/main/charts/rekor
+
 maintainers:
   - name: The Sigstore Authors
+    url: https://sigstore.dev/
 
 dependencies:
   - name: trillian
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -6,8 +6,8 @@ description: |
 
 type: application
 
-version: 1.0.0-rc.2
-appVersion: 1.0.0-rc.2
+version: 1.0.0-rc.3
+appVersion: v1.0.0-rc.1
 
 keywords:
   - security
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -36,6 +36,6 @@ annotations:
     - name: curlimages/curl
       image: docker.io/curlimages/curl@sha256:dca6e1b1c8e7b8b8e7be4e79fc78a858d12fd56245cb31bfa281dbf7c73a6498
     - name: rekor-server
-      image: gcr.io/projectsigstore/rekor-server:v1.0.0-rc.1@sha256:eb351dc8daa63599b41406c4a7eafff6617adf9fc604818349159ccbb60e613a
+      image: gcr.io/projectsigstore/rekor-server@sha256:eb351dc8daa63599b41406c4a7eafff6617adf9fc604818349159ccbb60e613a
     - name: redis
       image: docker.io/redis@sha256:6c42cce2871e8dc5fb3e843ed5c4e7939d312faf5e53ff0ff4ca955a7e0b2b39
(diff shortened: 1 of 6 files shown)


charts/rekor/Chart.yaml
```diff
@@ -14,12 +16,15 @@ keywords:
...
+  - https://github.com/sigstore/helm-charts/tree/main/charts/rekor
+
 maintainers:
   - name: The Sigstore Authors
+    url: https://sigstore.dev/
 
 dependencies:
   - name: trillian
-    version: 0.1.8
+    version: 0.1.10
```
marcofranssen, line 27: Requires #297 to be merged first

charts/rekor/values.yaml
```diff
@@ -62,8 +62,8 @@ server:
     registry: gcr.io
     repository: projectsigstore/rekor-server
     pullPolicy: IfNotPresent
-    # crane digest gcr.io/projectsigstore/rekor-server:v0.12.0
-    version: "sha256:a72892d8d663624a89f7e7cb1113a0b38caf794a32d7e6ddc178f8e8aac2c8da"
+    # -- v0.12.0
```
priyawadhwa, line 65: could we live the `crane digest` comment as a hint for anyone trying to update this chart later?

charts/rekor/values.yaml
```diff
@@ -26,8 +26,8 @@ redis:
     registry: docker.io
     repository: redis
     pullPolicy: IfNotPresent
-    # -- 6.2.6-alpine3.15
-    version: "sha256:6c42cce2871e8dc5fb3e843ed5c4e7939d312faf5e53ff0ff4ca955a7e0b2b39"
+    # -- 7.0.4-alpine3.16
+    version: sha256:dc1b954f5a1db78e31b8870966294d2f93fa8a7fba5c1337a1ce4ec55f311bc3
```
priyawadhwa, line 30: i haven't updated these images before, but it might make sense to update each image version separately, just in case the release fails we know what went wrong 🤔 

@cpanato what do you usually do here?

charts/rekor/values.yaml
cpanato: yes, i did a request before to keep that

charts/rekor/values.yaml
cpanato: yes, can make incremental, it is just a bit more work, only if that is good to be bumped, with i dont know since that is a major bump from 6 to 7 maybe there is a breaking change

charts/rekor/values.yaml
marcofranssen: That command is now documented in the CONTRIBUTING.md

charts/rekor/values.yaml
marcofranssen: Only bumped to newer patch version now.

charts/rekor/values.yaml
marcofranssen: I have added the `crane digest` hints in all places now. Also in the places it wasn't there yet.

priyawadhwa (dismissed):

charts/rekor/Chart.yaml
```diff
@@ -1,10 +1,12 @@
...
 name: rekor
-description: Part of the sigstore project, Rekor is a timestamping server and transparency log for storing signatures, as well as an API based server for validation
+description: |
+  Part of the Sigstore project, Rekor is a timestamping server and transparency
+  log for storing signatures, as well as an API based server for validation.
 
 type: application
 
-version: 1.0.0-rc.1
+version: 1.0.0-rc.2
```
priyawadhwa, line 9: @cpanato is this the recommended way to update this version, since we might still have to relase an rc.2 in rekor later?

marcofranssen: I have enabled edits by maintainers on this PR. Also wan't sure on this. But I think there will be anyhow a point where the version and appVersion will live their own lives at some point.

Anyhow, feel free to push as needed on the branch, to get this PR finished.

Don't forget to run the `helm-docs charts/rekort` to update the README.md as well.

cpanato: maybe we can wait till we have the rekor 1.0.0 out

marcofranssen: Shouldn't we have the Chart.lock in there prior to a 1.0.0 release?

cpanato: agree, but I would drop the other images updates for now, or do it in a follow up PR, want to reduce any possible issues when we rollout at this stage

marcofranssen: Ok I can make that adjustment.  Should be matter of dropping the commit if I used my normal hygiene when committing 😊

marcofranssen: @cpanato done. Also fixed some errors as pointed out in the files. Those where introduced in other PRs being merged.

PR #307, opened 2022-09-22
Add chart for TUF

flxw:
## Description of the change

As discussed with @sabre1041, this PR adds a chart for TUF _and_ integrates it with scaffold.

None that I could find

## How to test:
1. Spin up a local cluster using kind
2. Install cert-manager: 
```
helm install \
  cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.8.0 \
  --set installCRDs=true
```
3. Install the scaffold helm chart from this branch, using the [values.test.yaml.txt](https://github.com/sigstore/helm-charts/files/9663561/values.test.yaml.txt) attached to this description
4. Wait for everything to come up
5. Create port-forwardings for tuf, rekor and fulcio to localhost
6. Extract root.json from the tuf secret and initialize cosign with it
7. Extract the ctlog public key and save it locally
8. Use the provided script to test the whole thing ([sign-image-locally.sh.txt](https://github.com/sigstore/helm-charts/files/9663548/sign-image-locally.sh.txt))
## Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). 
Where applicable, update and bump the versions in any associated umbrella chart
- [ ] JSON Schema generated.


charts/scaffold/templates/copy-secrets-job.yaml
```diff
@@ -0,0 +1,64 @@
+{{- if .Values.copySecretJob.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Values.tuf.forceNamespace }}
```
sabre1041, line 5: For the `namespace` field, take a look how the other charts leverage a partial called `<chart>.namespace`. This capability is one of the components that will be added to a common library chart in the future

charts/tuf/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
+name: tuf
+description: A framework for securing software update systems - the scaffolding implementation
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
```
sabre1041, line 6: What does `1.16.0` refer to?

charts/tuf/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+home: https://sigstore.dev/
+sources:
+  - https://github.com/sigstore/scaffolding
+  - https://github.com/sigstore/helm-charts/tree/main/charts/tuf
```
sabre1041, line 11: The 2nd source can be removed

charts/tuf/templates/ingress.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - host: {{ required "An Ingress hostname is required" .host | quote }}
+      http:
+        paths:
+          - path: {{ .path }}
+            pathType: {{ default "Prefix" .pathType }}
+            backend:
+              service:
+                name: {{ $.Values.service.name }}
+                port:
+                  number: {{ $.Values.deployment.port | int }}
```
sabre1041, line 23: `deployment.port` refers to port `8080` which is the target port. Port `80` is the frontend for the service

charts/tuf/values.yaml
```diff
@@ -0,0 +1,41 @@
+forceNamespace: ""
+imagePullSecrets: []
+serviceAccountName: tuf-sa
```
sabre1041, line 3: Can this be changed to `tuf` as its already implied that it is an SA?

sabre1041 (changes requested):
This is REALLY good. Had to make a few tweaks that are called out in the comments. We should be good to integrate soon!

charts/scaffold/README.md
```diff
@@ -1,55 +1,45 @@
...
-This command deploys the default configuration for the scaffold chart. The [Parameters] section describes the various ways in which the chart can be configured.
-
-## Uninstallation
-
-```shell
-helm uninstall [RELEASE_NAME]
-```
-
-The previous command removes the previously installed chart.
+| https://sigstore.github.io/helm-charts | tuf | 0.1.0 |
```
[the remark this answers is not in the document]
flxw: Yes, this is what `helm-docs` put out. I didn't re-add the sections because they didn't carry a lot of value imho (installing a chart is a simple command). I brought back these sections in  329d7b4.

charts/tuf/Chart.yaml
flxw: Good catch! When I wrote this number, I know had some logic behind it.
Unfortunately, I can't quite get it back together.
I changed the number to match the version of the tuf image that is used here:
https://github.com/sigstore/scaffolding/pkgs/container/scaffolding%2Fserver/38864075?tag=v0.4.6
0a2b3df

charts/tuf/Chart.yaml
flxw: Thanks! Addressed in c8784c9

charts/tuf/values.yaml
flxw: Yes. Addressed in 22ad794

haydentherapper:
cc @asraa

asraa:
Out of curiousity, did the CTFE key not get detected automatically? It should pick up the key from TUF rather than the env var in the script I see

charts/tuf/templates/ingress.yaml
flxw: Addressed in fcca29d

charts/scaffold/templates/copy-secrets-job.yaml
flxw: Yes, this is an interesting one. Adressed for TUF in c5c116d.
If we want to do it properly, we should use a similar macro here for all the other namespaces where the secret is copied from, like rekor, fulcio and ctlog.

asraa:
> How would I check that? Could you elaborate what you mean by "it should pick up the key from TUF"?

I think it should just work.
You've defined 
```
| tuf.secrets.ctlog.name | string | `"ctlog-public-key"` |  |
| tuf.secrets.ctlog.path | string | `"ctlog-pubkey"` |  |
```

and sigstore's TUF client picks up the CT pubkey by name: https://github.com/sigstore/cosign/blob/31e665415f2da47356e4657e751443fcf5f394ed/cmd/cosign/cli/fulcio/fulcioverifier/ctl/verify.go#L37

which is written by scaffolding here:
https://github.com/sigstore/scaffolding/blob/83efd5f53bb7bf99777729a17374def24e9087b5/cmd/tuf/createsecret/main.go#L43

so I think all I'm asking was: is it necessary to have defined 
`export SIGSTORE_CT_LOG_PUBLIC_KEY_FILE=$HOME/celonis/sigstore/scaffolding-test/ctfe.pub`
in your test script, or if you unset that var, do things work fine after running `./cosign initialize` in step (6) of your test

sabre1041 (changes requested):
@flxw It looks like the recent changes caused a regression and the scaffolding no longer functions correctly

charts/scaffold/templates/_helpers.tpl
```diff
@@ -0,0 +1,22 @@
...
+
+
+{{/*
+Define the raw tuf.namespace template if set with forceNamespace or .Release.Namespace is set
+*/}}
+{{- define "tuf.rawnamespace" -}}
+{{- if .Values.tuf.forceNamespace -}}
+{{ print .Values.tuf.forceNamespace }}
+{{- else -}}
+{{ print .Release.tuf.Namespace }}
```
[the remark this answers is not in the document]
flxw: Thanks for noticing, that wasn't an easy one to pin down.
It's fixed in 22ef7a7
If you are curious, it had to do with calling a nested template with the correct context: https://stackoverflow.com/questions/47791971/how-can-you-call-a-helm-helper-template-from-a-subchart-with-the-correct-conte

flxw:
> 

Hey Asra, it took me a bit to decipher the question :D
I think there's two parts. Part one, regarding the naming of the inputs:

asraa:
> According to my test experiences, the env var needs to be set so that cosign does not fall back to the official sigstore pubkey.

That is really strange, and would indicate that the TUF server from sigstore/scaffolding is not signing off on the `ctfe.pub`.... but with the links above it does. 

> cosign does not fall back to the official sigstore pubkey.

After cosign initialize with the scaffolding root, cosign does not fallback to official sigstore pubkey. Maybe I can try the steps to reproduce the error myself

flxw:
> 

I just reran it on my machine and it seems I was wrong and you're right. The CTFE env entry is not needed.

asraa (dismissed):

sabre1041 (changes requested):
@flxw I worked around the conflicts manually and confirmed the new capabilities work as expected. 

As soon as the conflicts are resolved, we should be good to merge.

flxw:
> @flxw I worked around the conflicts manually and confirmed the new capabilities work as expected.
> 
> As soon as the conflicts are resolved, we should be good to merge.

Thanks! I resolved the merge conflicts now. Quite the commit history now, would you like me to rebase it?

sabre1041:
@flxw sure!

sabre1041:
@flxw would you be able to sign the DCO?

flxw:
> @flxw would you be able to sign the DCO?

Signed, squashed and clean :)

PR #315, opened 2022-09-27
Add job extra annotations

cpanato:
## Description of the change
- add createtree job annotation

that will allow us to add job annotations, and in the case of sigstore to add argocd annotations to address the issues described here https://github.com/sigstore/public-good-instance/issues/782

partially address https://github.com/sigstore/public-good-instance/issues/782

after we merge this will update argocd side

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.3.18
+version: 0.3.19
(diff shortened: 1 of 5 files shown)


k4leung4 (approved):

hectorj2f (approved):

vaikas:
This is great! I think we should do this for all the jobs that we use, but those can be done later.

cpanato:
> This is great! I think we should do this for all the jobs that we use, but those can be done later.

yes, will do, start with this one :)

PR #330, opened 2022-09-30
add --log_type flag to fulcio to enable production logging

priyawadhwa:
we need this to close out one of the remaining items in https://github.com/sigstore/public-good-instance/issues/387

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/fulcio/templates/fulcio-deployment.yaml
+++ b/charts/fulcio/templates/fulcio-deployment.yaml
@@ -51,6 +51,9 @@ spec:
             - "--fileca-key-passwd"
             - "$(PASSWORD)"
             {{- end }}
+            {{- if .Values.server.logging.production }}
+            - "--log_type=prod"
+            {{- end -}}
             {{- if eq .Values.server.args.certificateAuthority "kmsca" }}
             - "--kms-resource={{ .Values.server.args.kms_resource }}"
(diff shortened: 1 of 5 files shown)


charts/fulcio/templates/fulcio-deployment.yaml
```diff
@@ -51,6 +51,9 @@ spec:
             - "--fileca-key-passwd"
             - "$(PASSWORD)"
             {{- end }}
+            {{- if .Values.server.logging.production }}
+            - "--log_type=prod"
```
k4leung4, line 55: should we pass the log_type rather than a boolean?
in the odd chance we need debugging logs?

k4leung4 (commented):
please update README.md and schema

charts/fulcio/templates/fulcio-deployment.yaml
priyawadhwa: there's only two types right now, dev (default) and prod. i copied this from the rekor chart from consistency, but can update both if you think that's better

k4leung4: dont feel strongly, fine as is for consistency

please update readme and schema with new value

priyawadhwa:
updated, thanks for the reminder!

k4leung4 (approved):
thank you

PR #331, opened 2022-09-30
Bump scaffold version to latest fulcio release

priyawadhwa:

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/scaffold/Chart.yaml
+++ b/charts/scaffold/Chart.yaml
@@ -4,7 +4,7 @@ description: Scaffolding the components of the sigstore architecture
 
 type: application
 
-version: 0.3.21
+version: 0.3.22
 
 keywords:
   - security
@@ -17,7 +17,7 @@ maintainers:
 
 dependencies:
   - name: fulcio
-    version: 0.4.12
+    version: 0.5.1
     repository: https://sigstore.github.io/helm-charts
     condition: fulcio.enabled
   - name: rekor


charts/scaffold/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Scaffolding the components of the sigstore architecture
 
 type: application
 
-version: 0.3.21
+version: 0.3.22
```
cpanato, line 7: if there is a breaking change we should bump here as well `0.4.0`

k4leung4: so fulcio helm chart 0.5.x was cut for fulcio app v0.6.x

looking at fulcio app v0.6.x, there doesnt seem to be any breaking change.
not sure what the reason fulcio was bumped to 0.6.x? https://github.com/sigstore/fulcio/releases/tag/v0.6.0

haydentherapper: "Change username format" was technically a breaking change - However, it's unlikely anyone was using this format, and we weren't using this in production. But it was a change in the format of a subject

priyawadhwa: Sounds like it's ok to make this upgrade. Could someone PTAL?

k4leung4 (approved):

charts/scaffold/Chart.yaml
k4leung4: approved

priyawadhwa: thanks kenny!

PR #340, opened 2022-10-07
Update rekor release to 1.0 prerelease

priyawadhwa:

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/rekor/Chart.yaml
+++ b/charts/rekor/Chart.yaml
@@ -4,7 +4,7 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.3.22
+version: 1.0.0-rc.0
 appVersion: 1.0.0-rc.1
 
 keywords:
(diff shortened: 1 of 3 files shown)


charts/rekor/Chart.yaml
```diff
@@ -4,8 +4,8 @@ description: Part of the sigstore project, Rekor is a timestamping server and tr
 
 type: application
 
-version: 0.3.21
-appVersion: 0.12.2
+version: 0.3.22
```
haydentherapper, line 7: Should we bump the version to 0.9.9 to signify a prerelease?

haydentherapper: Just keeping adding 9s if we have more prereleases :)

hectorj2f: I had a look and I believe we could create a Chart with `version: 1.0.0-rc.0`, cc @cpanato

cpanato: yes we should be able to do that

priyawadhwa: thanks, updated!

hectorj2f (approved):

PR #341, opened 2022-10-07
Update Fulcio helm chart for prerelease

haydentherapper:
Signed-off-by: Hayden Blauzvern <hblauzvern@google.com>

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 0.9.9
+version: 1.0.0-rc.0
 appVersion: 1.0.0-rc.0
 
 keywords:
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -5,8 +5,8 @@ description: |
 
 type: application
 
-version: 0.5.2
-appVersion: 0.6.0
+version: 0.9.9
+appVersion: 1.0.0-rc.0
 
 keywords:
   - security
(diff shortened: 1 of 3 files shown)


charts/fulcio/Chart.yaml
```diff
@@ -5,8 +5,8 @@ description: |
 
 type: application
 
-version: 0.5.2
-appVersion: 0.6.0
+version: 0.9.9
```
haydentherapper, line 8: Happy to change if there's a preference, I just thought we should signify this is approaching 1.0

hectorj2f: It is a bit risky in case we need to update the chart with a new release candidate in case of bugs before 1.0

haydentherapper: I was thinking we could just add .9 again for each rc. Happy to revert this back though and just increment the minor version until 1.0, no strong preference!

hectorj2f: It is okay with me, just wanted to know how we'll deal with more RCS.

hectorj2f (dismissed):

charts/fulcio/Chart.yaml
hectorj2f, line 8: I had a look and I believe we could create a Chart with `version: 1.0.0-rc.0`, cc @cpanato

haydentherapper: Updated!

hectorj2f (approved):

PR #374, opened 2022-10-19
Fix the secret mounting to work without specifying keys.

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

## Description of the change

Previously each key in the secret for ctlog was explicitly specified. But in order to support multiple Fulcios, they were moved to `fulcio-XXX` where XXX specifies the index we mount everything from the secret.
Also, we always write the config in secret, so there's no need to mount the config map anymore.

Relevant PRs from scaffolding:
This in particular:
https://github.com/sigstore/scaffolding/pull/352/files#diff-dc1ec6e1748d93e1822f7b79815b687dc9e2effb36adfc168dc1354b740164c0

https://github.com/sigstore/scaffolding/pull/334/files
https://github.com/sigstore/scaffolding/pull/341/files

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/ctlog/values.yaml
+++ b/charts/ctlog/values.yaml
@@ -5,8 +5,6 @@ namespace:
 server:
   replicaCount: 1
   config:
-    secret:
-      name: ctlog-config
     key: treeID
     treeID: ""
   image:
--- a/charts/ctlog/values.yaml
+++ b/charts/ctlog/values.yaml
@@ -6,7 +6,6 @@ server:
   replicaCount: 1
   config:
     secret:
-      enabled: false
       name: ctlog-config
     key: treeID
     treeID: ""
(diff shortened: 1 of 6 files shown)


charts/ctlog/values.yaml
```diff
@@ -6,7 +6,6 @@ server:
   replicaCount: 1
   config:
     secret:
-      enabled: false
       name: ctlog-config
```
k4leung4, line 9: can we delete this as well, it does not seem to be used anywhere, which means the entire `secret` can be deleted

cpanato (dismissed):

k4leung4 (approved):
does this depend on a new scaffolding image/release?

vaikas:
> does this depend on a new scaffolding image/release?

You mean the ctlog? There were no changes there, just instead of putting the keys and config in different directories, they are now put into the same directory. Also teh --config flag points to difference file now because it's mounted directly as the key name.
does that make sense?

PR #377, opened 2022-10-21
Exclude scaffolding for TUF

sabre1041:
Excluding scaffolding so that TUF can release.

Subsequent PR to be created to support reenabling scaffolding

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -5,3 +5,5 @@ remote: origin
 target-branch: main
 chart-repos:
   - sigstore=https://sigstore.github.io/helm-charts
+excluded-charts:
+  - scaffolding


flxw (approved):

hectorj2f:
@sabre1041 Could you add more context why you need to exclude scaffolding?

k4leung4:
same as @hectorj2f 

I don't understand this change, context would be greatly appreciated
thank you

sabre1041:
@hectorj2f @k4leung4 the issue is that the tuf chart didnt release successfully.we need to release tuf first by excluding scaffolding and then reenable scaffolding so that it can release the newly available tuf release

this was caused by a PR that included changes for both tuf and scaffolding

k4leung4 (approved):

PR #380, opened 2022-10-24
Configure fulcio secret name

flxw:
## Description of the change

This adds an explicit option for setting the name of the secret that's used by the Fulcio deployment.
Additionally, a change is introduced to actually use the `createcerts.enabled` flag.
Before, both settings were implicit.

## Checklist

- [x] JSON Schema generated.


charts/fulcio/Chart.yaml
```diff
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 1.0.0
+version: 1.1.0
```
k4leung4, line 8: I would bump this to 2.x.x and mention this release will be a breaking change to existing users.

k4leung4 (changes requested):

charts/fulcio/Chart.yaml
flxw: Thanks for the quick feedback!
You're right. I didn't consider the breaking consequences of changing the if condition.
I pushed the change just now.

k4leung4 (dismissed):
thanks for the change

charts/fulcio/Chart.yaml
```diff
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 1.0.0
+version: 2.0.0
```
cpanato, line 8: you bump to 2.0.0 there is any breaking change?
if yes can you write those down and maybe how users can upgrade existing deployment

thanks!

sabre1041 (commented):
question regarding the chart version

charts/fulcio/README.md
```diff
@@ -2,7 +2,7 @@
 
 <!-- This README.md is generated. Please edit README.md.gotmpl -->
 
-![Version: 1.0.0-rc.1](https://img.shields.io/badge/Version-1.0.0--rc.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.0.0-rc.1](https://img.shields.io/badge/AppVersion-1.0.0--rc.1-informational?style=flat-square)
+![Version: 1.0.0](https://img.shields.io/badge/Version-1.0.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.0.0](https://img.shields.io/badge/AppVersion-1.0.0-informational?style=flat-square)
```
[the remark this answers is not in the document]
flxw: It was k4leung4's suggestion, and it is indeed a breaking change if you think about it.
See the comment: https://github.com/sigstore/helm-charts/pull/380#discussion_r1003400472

I missed the badge link here, happy to update it either way. Do you disagree with bumping to 2.0.0?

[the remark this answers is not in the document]
sabre1041: lets just update the `README.md` and move forward with `2.0.0`

sabre1041 (approved):
LGTM

PR #385, opened 2022-10-25
Add tuf secrets

flxw:
## Description of the change

Adds a feature to create the required secrets for TUF as part of the deployment.
The required secrets were missing during the automated tests.

## Checklist

- [x] JSON Schema generated.


charts/tuf/values.yaml
```diff
@@ -15,15 +15,19 @@ deployment:
   port: 8080
 
 secrets:
+  create: true
```
sabre1041, line 18: Lets default to `false` and then provide testing values so that we can execute the logic that would be applied if necessary

sabre1041 (changes requested):

charts/tuf/values.yaml
flxw: Done :)

k4leung4 (commented):
looks good,
please update schema

charts/tuf/values.yaml
```diff
@@ -15,15 +15,22 @@ deployment:
   port: 8080
 
 secrets:
+  create: false
   rekor:
     name: rekor-public-key
+    key: public
     path: rekor-pubkey
+    value: b3B0aW9uYWw=
```
hectorj2f, line 23: Q: I assume you don't need to set the value here.

flxw: Yup, changed after discussion with Andrew

sabre1041 (approved):

sabre1041:
LGTM. There are a few further enhancements that can be applied. But as of this point, this should allow for reenabling full functionality

sabre1041:
@hectorj2f @k4leung4 can you review once more so we can press forward?

k4leung4 (approved):

hectorj2f (approved):

PR #401, opened 2022-11-04
Run Rekor with rekor_signer.signer set to a file

flxw:
## Description of the change

This PR adds capabilities to run Rekor with `--rekor_server.signer=/path/to/a/private/key/file`.
The signer could already be set, but it was not possible to feed in the private key file to the specified path through a secret. The PR adds that capability.

To test it, generate a private key and create a secret with it:
```
openssl ecparam -name prime256v1 -genkey -noout -out key.pem
kubectl create secret generic rekor-private-key --from-file=private=key.pem
```

Afterwards, set the following values:
```
rekor:
  signer: /key/private # for file-based secrets this needs to be secretMountPath + privateKeySecretKey
  signerFileSecretOptions:
    secretName: rekor-private-key
    secretMountPath: /key
    privateKeySecretKey: private
```

The change is backwards-compatible.

Two issues that I can see:

1. There still isn't any capability to pass in the password for the private key.
Would this be needed for both GCP as well as file-based signer private keys?
My understanding of GCP is limited, so I'm happy about any expertise there!
2. The values are a bit unclean in my opinion. I didn't implement it because it would be a breaking change. However, I'd prefer to have the following structure:
```
rekor:
  signer:
    arg: /key/private # for file-based secrets this needs to be secretMountPath + privateKeySecretKey
    secretName: rekor-private-key
    secretMountPath: /key
    privateKeySecretKey: private
```

## Additional Information

 
The signer implementation in Rekor is very basic, and looks like an update could happen there soon:
https://github.com/sigstore/rekor/blob/main/pkg/signer/signer.go#L27

## Checklist

- [x] JSON Schema generated.


JAORMX (dismissed):

k4leung4 (approved):

charts/rekor/templates/server/deployment.yaml
```diff
@@ -121,3 +126,10 @@ spec:
             claimName: {{ template "rekor.server.fullname" . }}
   {{- end }}
 {{- end }}
+{{- if .Values.server.signerFileSecretOptions }}
+        - name: {{ .Values.server.signerFileSecretOptions.secretName }}-volume
+          secret:
+            secretName: {{ .Values.server.signerFileSecretOptions.secretName }}
+            items:
+            - key: {{ .Values.server.signerFileSecretOptions.privateKeySecretKey }}
```
sabre1041, line 134: `path` is also required

sabre1041 (changes requested):
Can `signerFileSecretOptions` be added with an empty map to the `values.yaml` so that users know where the struct should be placed?

Also added another comment to address related to missing field

charts/rekor/templates/server/deployment.yaml
flxw: Good point, thanks for mentioning that detail. Added that in [e80acac](https://github.com/sigstore/helm-charts/pull/401/commits/e80acac2c4a57c237e0abda83cd92dd0f79bba61)

charts/rekor/templates/server/deployment.yaml
```diff
@@ -69,6 +69,11 @@ spec:
             - name: storage
               mountPath: {{ template "rekor.server.fileAttestationStorage.path" . }}
 {{- end -}}
+{{- if .Values.server.signerFileSecretOptions }}
+            - name: {{ .Values.server.signerFileSecretOptions.secretName }}-volume
+              mountPath: {{ .Values.server.signerFileSecretOptions.secretMountPath }}                                                                                                                      │
```
sabre1041, line 74:
```suggestion
              mountPath: {{ .Values.server.signerFileSecretOptions.secretMountPath }}
```

sabre1041 (changes requested):

sabre1041 (approved):
LGTM

sharadhirao:
Hi @sabre1041,
I have created a private key and created a secret like this:
```
openssl ecparam -name prime256v1 -genkey -noout -out key.pem
kubectl create secret generic rekor-private-key --from-file=private=key.pem
```

Afterwards, set the following values:
```
rekor:
  signer: /key/private # for file-based secrets this needs to be secretMountPath + privateKeySecretKey
  signerFileSecretOptions:
    secretName: rekor-private-key
    secretMountPath: /key
    privateKeySecretKey: private
```

It says `secretMountSubPath` is required for this. Can you tell me what should I have as a `secretMountSubPath` in this setup? It would be a great help. I did not find anything in values.yaml

PR #414, opened 2022-11-14
Trillian: Add ability to specify custom storage system

JAORMX:
## Description of the change

Trillian: Add ability to specify custom storage system

Currently, the Trillian installation is hard-coded to use the mysql
storage system. While this is mostly fine (it is the most common and
used storage system for Trillian after all!), this makes it non trivial
to try custom backends for Trillian.

There is work on the way to introduce [a CockroachDB storage system for
Trillian](https://github.com/google/trillian/pull/2834). And, while this
is currently not usable using the mainline Trillian image, it is
possible to try out if one specifies a custom image for trillian.

To enable this, some variables were introduced:

* `storageSystem.driver`: defines the storage backend to use for both the log
  signer and the log server. Defaults to `mysql`.

* `storageSystem.envCredentials`: is the kubernetes deployment
  definition of environment variables to aide the log signer in connecting to the
  desired database. When set to `null` it defaults to the environment
  variables used by MySQL:

A values file that would take this work into use would look as follows:

```yaml
storageSystem:
  driver: crdb
  envCredentials:
    - name: CRDB_USER
      valueFrom:
        secretKeyRef:
          name: mySecret
          key: crdb-user
    - name: CRDB_PASSWORD
      valueFrom:
        secretKeyRef:
          name: mySecret
          key: crdb-password
    - name: CRDB_HOST
      value: crdb.trillian.svc.cluster.local
logServer:
  image:
    registry: ghcr.io
    repository: equinixmetal-security/trillian-log-server
    version: latest
  extraArgs:
    - "--crdb-uri=postgres://$(CRDB_USER):$(CRDB_PASSWORD)@$(CRDB_HOST):26257/"
logSigner:
  image:
    registry: ghcr.io
    repository: equinixmetal-security/trillian-log-signer
    version: latest
  extraArgs:
    - "--crdb-uri=postgres://$(CRDB_USER):$(CRDB_PASSWORD)@$(CRDB_HOST):26257/"
mysql:
  enabled: false
```

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/trillian/templates/trillian-log-server/deployment.yaml
+++ b/charts/trillian/templates/trillian-log-server/deployment.yaml
@@ -68,25 +68,7 @@ spec:
           args:
 {{  include "trillian.logServer.args" . | indent 12 }}
           env:
-            - name: MYSQL_USER
-              valueFrom:
-                secretKeyRef:
-                  name: {{ template "mysql.secretName" . }}
-                  key: mysql-user
-            - name: MYSQL_PASSWORD
-              valueFrom:
-                secretKeyRef:
-                  name: {{ template "mysql.secretName" . }}
-                  key: mysql-password
-            - name: MYSQL_DATABASE
-              valueFrom:
-                secretKeyRef:
-                  name: {{ template "mysql.secretName" . }}
-                  key: mysql-database
-            - name: MYSQL_HOSTNAME
-              value: {{ template "mysql.hostname" . }}
-            - name: MYSQL_PORT
-              value: {{ .Values.mysql.port | quote }}
+{{- include "trillian.storageSystem.envCredentials" . | indent 12}}
           ports:
 {{- include "trillian.containerPorts" .Values.logServer.service.ports | indent 12 }}
 {{- if .Values.logServer.livenessProbe }}
--- a/charts/trillian/values.yaml
+++ b/charts/trillian/values.yaml
@@ -15,6 +15,9 @@ initContainerImage:
     version: "sha256:7d921b6d368fb1736cb0832c6f57e426c161593c075847af3378eb3185801cea"
     imagePullPolicy: IfNotPresent
 
+storageSystem:
+  driver: mysql
+  envCredentials: null
 mysql:
   gcp:
     enabled: false
--- a/charts/trillian/Chart.yaml
+++ b/charts/trillian/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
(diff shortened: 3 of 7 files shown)


charts/trillian/values.yaml
```diff
@@ -15,6 +15,7 @@ initContainerImage:
     version: "sha256:7d921b6d368fb1736cb0832c6f57e426c161593c075847af3378eb3185801cea"
     imagePullPolicy: IfNotPresent
 
+storageSystem: mysql
```
k4leung4, line 18: we should update `values.schema.json` with the new fields

charts/trillian/templates/trillian-log-server/deployment.yaml
```diff
@@ -68,6 +68,7 @@ spec:
           args:
 {{  include "trillian.logServer.args" . | indent 12 }}
           env:
```
k4leung4, line 70: since logServer and logSigner have the same `env` args, does it make sense to hoist it into _helpers.tpl? to avoid duplication.

charts/trillian/templates/trillian-log-server/deployment.yaml
```diff
@@ -87,6 +88,11 @@ spec:
               value: {{ template "mysql.hostname" . }}
             - name: MYSQL_PORT
               value: {{ .Values.mysql.port | quote }}
+{{- else if .Values.logServer.storageSystemEnvCredentials }}
```
k4leung4, line 91: is the intent to migrate the mysql usage to use `storageSystemEnvCredentials` as well?

JAORMX: I hadn't thought about this. It could certainly be an option!

charts/trillian/templates/trillian-log-server/deployment.yaml
JAORMX: sure, I can do that.

JAORMX:
@k4leung4 I addressed the comments. Thanks for the review!

k4leung4 (commented):
looks good. thanks for the changes.

please bump the helm chart version 
https://github.com/sigstore/helm-charts/blob/main/charts/trillian/Chart.yaml#L8

JAORMX:
> looks good. thanks for the changes.
> 
> please bump the helm chart version https://github.com/sigstore/helm-charts/blob/main/charts/trillian/Chart.yaml#L8

done 😄

PR #418, opened 2022-11-17
Add createctconfig.secret option to ctlog helm chart

priyawadhwa:

cc @pwelch @vaikas I think just putting the secret option under createctconfig might be easier, but wanted to make sure it'll work with the different shards?

### DIFF
--- a/charts/ctlog/Chart.yaml
+++ b/charts/ctlog/Chart.yaml
@@ -4,7 +4,7 @@ description: Certificate Log
 
 type: application
 
-version: 0.2.34
+version: 0.2.35
 appVersion: 0.3.0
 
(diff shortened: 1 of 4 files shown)


vaikas:
Yes, as long as we make sure to match these lines so that creation of a shard config and using it are consistent:
https://github.com/sigstore/public-good-instance/blob/main/argocd/utilities/templates/ctlog-shards.yaml#L55
And then maybe setting this new variable to match it here (well, add it here and make it match):
https://github.com/sigstore/public-good-instance/blob/main/argocd/utilities/templates/ctlog-shards.yaml#L68

vaikas (approved):

PR #419, opened 2022-11-17
Fix template error in rekor chart

priyawadhwa:
The ingress host name is required, so we should have a default value for the name since we have a default ingress rule.


vaikas (approved):
Woot Woot, thanks!

PR #421, opened 2022-11-18
policy-controller: Make webhook namespace selector configurable

JAORMX:
## Description of the change

This allows for more flexibility when setting up the namespace selector.
e.g. folks are able to set up whatever configuration they see fit for
their deployments.

This introduces two new components to the helmm chart:

* A `policy-controller.webhook.namespaceSelector` template which may be
  leveraged as an `include` statement.

* A new key with the path `webhook.namespaceSelector` which allows for
  configuring the aforementioned namespace selector.

The default is left as it was before, this way it won't affect new
deployments and upgrades will be seamless.

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/policy-controller/Chart.yaml
+++ b/charts/policy-controller/Chart.yaml
@@ -8,7 +8,7 @@ sources:
 type: application
 
 name: policy-controller
-version: 0.3.6
+version: 0.3.7
 appVersion: 0.4.2
 
 maintainers:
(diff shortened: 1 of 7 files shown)


hectorj2f (commented):
Please, add the new field in the README.

JAORMX:
@hectorj2f done, thanks for the review!

hectorj2f (approved):

PR #461, opened 2023-01-17
Configure fulcio resources

vipulagarwal:
## Description of the change

Adding ability to configure k8s resource requests and limits in all the components of fulcio.

## Additional Information

At present, we cannot configure resource requests and limits for some components like one time jobs, initContainers and some deployments via Sigstore helm charts. This becomes a problem when `ResourceQuota` is defined for namespaces and prevents pods to spin up. It is a best practice to define resource requests and limits for all k8s resources.

 

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/fulcio/README.md
+++ b/charts/fulcio/README.md
@@ -86,6 +86,7 @@ helm uninstall [RELEASE_NAME]
 | createcerts.image.version | string | `"sha256:73e7ac35d0e5169bd14a5cb6caed2e7d44277dec3d1de92e08f4d055523089a1"` |  |
 | createcerts.name | string | `"createcerts"` |  |
 | createcerts.replicaCount | int | `1` |  |
+| createcerts.resources | string | `""` |  |
 | createcerts.securityContext.runAsNonRoot | bool | `true` |  |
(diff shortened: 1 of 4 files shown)


JAORMX (approved):

hectorj2f (approved):

charts/fulcio/README.md
```diff
@@ -86,6 +86,7 @@ helm uninstall [RELEASE_NAME]
 | createcerts.image.version | string | `"sha256:73e7ac35d0e5169bd14a5cb6caed2e7d44277dec3d1de92e08f4d055523089a1"` |  |
 | createcerts.name | string | `"createcerts"` |  |
 | createcerts.replicaCount | int | `1` |  |
+| createcerts.resources | string | `""` |  |
```
k4leung4, line 89: resources isnt a string, should this be an object?

cmurphy:
The question about resources being a string was never responded to. This would need some more work to be able to go in, apart from rebasing this would need to be properly added to values.yaml and the values schema and the README needs to be generated, not manually edited. I'm going to close this but feel free to reopen if this is something you still want to work on.

PR #477, opened 2023-02-28
FEAT#Allow Enable/Disable Kubernetes Service Links

nkreiger:
## Description of the change

This Change/Proposal, is to allow setting the enableServiceLinks boolean in a kubernetes deployment via the helm charts. This is kind of an obscure configuration ask.

This variable (true by default), will automatically map all the service names and hosts into the environment variables of a kubernetes deployment. However, there is a limit to this, and when that limit is hit, you get the error `too many args`, whenever a pod restarts or during a new deployment.

Honestly, not sure why the default is true, it should only effect you if you are not using the DNS name when interacting with other services, which is not the case here.

## Checklist

- [ ] JSON Schema generated.


nkreiger:
I wasn't sure if I should just keep this forked, or open a PR, if this is something to move forward with, I can do the checklist, let me know.

sabre1041 (changes requested):
A few items that need to be added before we can look to integrate this feature:

* Instead of a dictionary called `serviceLinks`, what about a boolean called `serviceLinksEnabled` and use that conditional and value in the deployment?
* `Chart.yaml` version bump
* JSON Schema values update
* Chart dependencies need to be updated between the charts

Given that this change is spread across multiple charts, separating them out across multiple PR's would be a better approach given the fact that chart testing needs to be able to resolve them effectively

nkreiger:
hey @sabre1041 yep, mostly considered this a request, more of the "final" pr. I will spread these out, and use the same strategy [here](https://github.com/bitnami/charts/commit/0f9a159c38b7a4ce693de79e243f6ee050275fdd) with the dictionary boolean

cmurphy:
This hasn't been updated in 3 years, I'm going to go ahead and close it but feel free to reopen it if this is something you still want to work on.

PR #482, opened 2023-03-07
Use multi-arch image for Trillian's netcat initContainer

codysoyland:
## Description of the change

This replaces `toolbelt/netcat` with the standard `alpine` multi-arch image, which makes this chart compatible with non-amd64 architectures. The alpine image includes the `nc` command.

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/trillian/README.md
+++ b/charts/trillian/README.md
@@ -47,6 +47,7 @@ helm uninstall [RELEASE_NAME]
 | createdb.image.repository | string | `"sigstore/scaffolding/createdb"` |  |
 | createdb.image.version | string | `"sha256:9aa98492115c465b0cecfd6dbb04411a40c0d2d7e5d7c510f5646bd1d825e3c7"` | v0.6.2 |
 | createdb.name | string | `"createdb"` |  |
+| createdb.resources | object | `{}` |  |
 | createdb.serviceAccount.annotations | object | `{}` |  |
 | createdb.serviceAccount.create | bool | `false` |  |
 | createdb.serviceAccount.name | string | `""` |  |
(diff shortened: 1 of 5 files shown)


charts/trillian/README.md
```diff
@@ -47,21 +47,19 @@ helm uninstall [RELEASE_NAME]
 | createdb.image.repository | string | `"sigstore/scaffolding/createdb"` |  |
 | createdb.image.version | string | `"sha256:9aa98492115c465b0cecfd6dbb04411a40c0d2d7e5d7c510f5646bd1d825e3c7"` | v0.6.2 |
 | createdb.name | string | `"createdb"` |  |
-| createdb.resources | string | `""` |  |
```
hectorj2f, line 50: I don't think we want to delete this fields.

sabre1041 (changes requested):
Can this reference be also updated

https://github.com/sigstore/helm-charts/blob/main/charts/trillian/Chart.yaml#L28-L29

sabre1041 (changes requested):
Can this reference be also updated

https://github.com/sigstore/helm-charts/blob/main/charts/trillian/Chart.yaml#L28-L29

charts/trillian/README.md
codysoyland: I'll try to update this this week... I'm not sure how that got deleted because I think I just used `helm-docs --dry-run` to generate it.

codysoyland: I checked and this didn't have a default in values.yaml, so I added it as the correct `map` type.

cmurphy:
@codysoyland is this still something you care about? Could you rebase?

bobcallaway:
#1232 will fix this

PR #520, opened 2023-05-02
Bugfix: Do not use ctlog.enabled to control ct-log-url flag

codysoyland:
## Description of the change

Context: I added the Helm value `ct_log_url` in https://github.com/sigstore/helm-charts/pull/509. It relied on `ctlog.enabled` to control the output of the flag. This unfortunately doesn't work with the scaffold Helm chart, which sets `fulcio.ctlog.enabled=false`, which causes the ct_log_url to be set to an empty string, which is not the intent with that chart. This PR changes the behavior so that:

- If `disable_ct_log` is set to true, the flag is output as a blank string
- If `ct_log_url` is set, the output is the value
- Otherwise, the old behavior is preserved (assume CT Log running in cluster)

The `disable_ct_log` setting is added here because `ct_log_url` may not be set to an empty string as it would fail the boolean test, and that's a feature I would like personally.

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/fulcio/values.schema.json
+++ b/charts/fulcio/values.schema.json
@@ -241,11 +241,21 @@
                                 "projects/test/locations/us-east1/caPools/test"
                             ]
                         },
+                        "disable_ct_log": {
+                            "type": "bool",
+                            "default": false,
+                            "title": "Disable CT log submission",
+                            "description": "If set, Fulcio will not submit issued certificates to a CT log.",
+                            "examples": [
+                                true,
+                                false
+                            ]
+                        },
                         "ct_log_url": {
                             "type": "string",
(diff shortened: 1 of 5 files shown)


charts/fulcio/values.schema.json
```diff
@@ -241,11 +241,21 @@
                                 "projects/test/locations/us-east1/caPools/test"
                             ]
                         },
+                        "disable_ct_log": {
+                            "type": "bool",
```
cpanato, line 245:
```suggestion
                            "type": "boolean",
```

cpanato (approved):

haydentherapper (approved):

PR #574, opened 2023-07-14
add parallel ingress support to ctlog chart

bobcallaway:
## Checklist

- [ ] JSON Schema generated.


haydentherapper (approved):
To confirm, we will have multiple “hosts” for each shard?

bobcallaway:
> To confirm, we will have multiple “hosts” for each shard?

no, one host, with multiple paths each pointing to a different backend service

bobcallaway:
this change is also slightly different than the changes for rekor/fulcio/dex in that ctlog is used as a "top level" chart for ctfe.sigstore.dev, as well as for ctlog-shards which calls scaffold in a loop (one ctlog instance per shard). GKE ingress however can only support one ingress object per external IP, so I implemented this differently so we could manually map the N shards into `paths` under the single ingress host in the "top level chart", but the k8s Service objects for the shards would still be created in the ctlog-shards loop as they are today.

PR #590, opened 2023-08-27
move to v2 of cloud-sql-proxy for trillian chart

bobcallaway:
moves to v2 version of https://github.com/GoogleCloudPlatform/cloud-sql-proxy for the trillian chart

adds support for using unix domain sockets to provide connectivity between containers instead of TCP

also adds support for other command line arguments, which enables environment specific features like structured logging, health checks, etc

### DIFF
--- a/charts/trillian/Chart.yaml
+++ b/charts/trillian/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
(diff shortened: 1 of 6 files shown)


cpanato (approved):
Thanks

haydentherapper (approved):

PR #649, opened 2023-10-31
set TUF_ROOT since homedir for nonroot image doesn't exist

bobcallaway:
## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/sigstore-prober/templates/deployment.yaml
+++ b/charts/sigstore-prober/templates/deployment.yaml
@@ -51,10 +51,15 @@ spec:
           runAsNonRoot: true
           runAsUser: 65534
           allowPrivilegeEscalation: false
-      {{- if .Values.spec.args.trustRekorAPIPublicKey }}
         env:
+          # TUF_ROOT set to /tmp since the home dir for user 65534 is set to /nonexistant
+        - name: TUF_ROOT
+          value: "/tmp"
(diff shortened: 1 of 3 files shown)


charts/sigstore-prober/templates/deployment.yaml
```diff
@@ -51,10 +51,15 @@ spec:
...
-      {{- if .Values.spec.args.trustRekorAPIPublicKey }}
         env:
+          # TUF_ROOT set to /tmp since the home dir for user 65534 is set to /nonexistant
+        - name: TUF_ROOT
+          value: "/tmp"
+      {{- if .Values.spec.args.trustRekorAPIPublicKey }}
         - name: SIGSTORE_REKOR_PUBLIC_KEY
           value: "/var/run/sigstore/public-key.pem"
+      {{- end }}
+      {{- if .Values.spec.args.trustRekorAPIPublicKey }}
```
haydentherapper, line 62: Do you need this extra condition, as it’s the same as the previous condition?

bobcallaway: yes, b/c it is wrapping a separate field. if we ever needed to make a change later, being explicit here is better IMO

haydentherapper: SG, just want to check it was intentional

haydentherapper (approved):

PR #651, opened 2023-11-01
Add signer password for file-based signers for Rekor

haydentherapper:
## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/rekor/README.md
+++ b/charts/rekor/README.md
@@ -1,6 +1,6 @@
 # rekor
 
-![Version: 1.3.12](https://img.shields.io/badge/Version-1.3.12-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.3.2](https://img.shields.io/badge/AppVersion-1.3.2-informational?style=flat-square)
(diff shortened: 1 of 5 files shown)


charts/rekor/README.md
```diff
@@ -169,6 +169,7 @@ Part of the sigstore project, Rekor is a timestamping server and transparency lo
 | server.sharding.filename | string | `"sharding-config.yaml"` |  |
 | server.sharding.mountPath | string | `"/sharding"` |  |
 | server.signer | string | `"memory"` |  |
+| server.signerPassword | string | `""` |  |
```
k4leung4, line 172: should we add a disclaimer this should only be used for testing?
passing password string here seems like bad practice in general

haydentherapper: Another option is to omit this entirely, as extraArgs can cover this. WDYT?

k4leung4: whats the motivation/use case for adding this arg?

extraArgs is also an option, especially it seems tedious to update the helm chart for every new flag

haydentherapper: Just had someone accessing about signing options for Rekor and I noticed this was missing. Have we formed guidance on when we should add a top-level argument vs using extraArgs?

haydentherapper: Thinking this over, I'm gonna close this out.

PR #711, opened 2024-02-06
Fix default key for tsa

priyawadhwa:
this should be the same as rekor [here](https://github.com/sigstore/helm-charts/blob/main/charts/scaffold/templates/copy-secrets-job.yaml#L46), i'll add it in for the tsa in that job as well in a follow up

### DIFF
--- a/charts/tuf/values.yaml
+++ b/charts/tuf/values.yaml
@@ -35,7 +35,7 @@ secrets:
   tsa:
     create: false
     name: tsa-cert-chain
-    key: cert-chain
+    key: key
     path: tsa.certchain.pem
 
 ingress:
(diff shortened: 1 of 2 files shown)


charts/tuf/values.yaml
```diff
@@ -35,7 +35,7 @@ secrets:
   tsa:
     create: false
     name: tsa-cert-chain
-    key: cert-chain
+    key: key
```
haydentherapper, line 38: Would it be possible to update the tsa chart to read cert-chain rather than key? That would be more accurate.

priyawadhwa: yep, addressed in https://github.com/sigstore/helm-charts/pull/712/files. i'll close this pr.

Issue #735, opened 2024-04-10
TSA secret always required for TUF deployment


xoanmi:
**Description**

Currently, the TSA secret is consistently mandated for [TUF deployment](https://github.com/sigstore/helm-charts/blob/main/charts/tuf/templates/deployment.yaml#L66-L70), despite its optional nature. Could consideration be given to implementing a flag that allows for the exclusion of this secret setup if deemed unnecessary?"

vipulagarwal:
There is some activity on #744 but if anyone is blocked by this can try the following:
```
  tsa:
    create: true
    name: tsa-cert-chain
    key: cert-chain
    path: dummyfile
    value: "dummyvalue"
```

This will make the `dummyfile` available in your tuf root though.

vipulagarwal:
@bobcallaway can we close this issue? It has been fixed in #744

PR #744, opened 2024-04-22
Tsa secret optional for tuf

cvegagimenez:
## Description of the change

Make the TSA secret reference optional for TUF chart.

[#735 ](https://github.com/sigstore/helm-charts/issues/735)

## Checklist

- [x] JSON Schema generated.

```Linting charts...

------------------------------------------------------------------------------------------------------------------------
 Charts to be processed:
------------------------------------------------------------------------------------------------------------------------
 tuf => (version: "0.1.13", path: "charts/tuf")
------------------------------------------------------------------------------------------------------------------------

"sigstore" already exists with the same configuration, skipping
Linting chart "tuf => (version: \"0.1.13\", path: \"charts/tuf\")"
Checking chart "tuf => (version: \"0.1.13\", path: \"charts/tuf\")" for a version bump...
Old chart version: 0.1.12
New chart version: 0.1.13
Chart version ok.
Validating ~/Workspace/helm-charts/charts/tuf/Chart.yaml...
Validation success! 👍

Linting chart with values file "charts/tuf/ci/ci-values.yaml"...

==> Linting charts/tuf
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed

------------------------------------------------------------------------------------------------------------------------
 ✔︎ tuf => (version: "0.1.13", path: "charts/tuf")
------------------------------------------------------------------------------------------------------------------------
All charts linted successfully```

### CODE REVIEW

File: charts/tuf/README.md [hunk 1]
```diff
@@ -54,6 +54,7 @@ A framework for securing software update systems - the scaffolding implementatio
 | secrets.rekor.name | string | `"rekor-public-key"` |  |
 | secrets.rekor.path | string | `"rekor.pub"` |  |
 | secrets.tsa.create | bool | `false` |  |
+| secrets.tsa.existingSecret | bool | `false` |  |
```
sabre1041, line 57: Why not make this `enabled` and set as `true`. If `secrets.tsa.create` is `true`, a new secret will be created. Otherwise, `secrets.tsa.name` is the name of the existing secret

cvegagimenez: I implemented the changes. I thought it would be better to set it as `enabled` as `false` by default since the Charts are independent, but I could change it if you consider the other way.

sabre1041: I like this approach a lot! It would be great if we can implement the same pattern across all of the secrets. Though that probably requires a separate PR to avoid encompassing too much into this issue. Unless you would want to rename the PR for that scope

cvegagimenez: As the TSA case, I set the default value to `false` to be independent.

cvegagimenez: @sabre1041 Any news on this?

sabre1041: @cvegagimenez syntactically this does work. However, in practice, for tuf to run properly, it will need at least one source of content (a secret) in order to start properly. Should we enforce that at least one secret is provided?

Also, would you be able to resolve the conflict in the `README.md` file?

cvegagimenez: Done. I also added the same checks for the other TUF objects.

sabre1041 (changes requested):
@cvegagimenez instead of omitting content when no secrets have been provided, an error should be thrown to ensure that the user provides at least one secret

sabre1041:
@cvegagimenez This looks good., However, while thinking it through in practice `enabled` should by default be `true` as it aligns with the current functionality of the chart. By setting to false, it would be a breaking change for anyone currently leveraging it.

The goal of this PR is to provide a way to opt out of providing secrets a, but in practice, this has now introduced the functionality where you have to opt in to achieve the current functionality. A simple swap of the default values as we should be good to integrate this change.

charts/tuf/templates/deployment.yaml
```diff
@@ -47,40 +47,9 @@ spec:
...
-    {{- end }}
-    {{- if .Values.deployment.tolerations }}
-      tolerations:
-{{ toYaml .Values.deployment.tolerations | indent 8 }}
-    {{- end }}
-    {{- if .Values.deployment.affinity }}
-      affinity:
-{{ toYaml .Values.deployment.affinity | indent 8 }}
-    {{- end }}
+    {{- end }}
```
cpanato, line 55: nit: new line

charts/tuf/templates/ingress.yaml
```diff
@@ -33,4 +33,4 @@ spec:
       secretName: {{ .secretName }}
     {{- end }}
 {{- end -}}
-{{- end -}}
+{{- end -}}
```
cpanato, line 36: nit: new line

charts/tuf/templates/_helpers.tpl
```diff
@@ -84,4 +84,52 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+        path: {{ .Values.secrets.rekor.path }}
+  {{- end }}
+  {{- if or (.Values.secrets.tsa.enabled) (.Values.secrets.tsa.create) }}
+  - secret:
+      name: {{ .Values.secrets.tsa.name }}
+      items:
+      - key: {{ .Values.secrets.tsa.key }}
+        path: {{ .Values.secrets.tsa.path }}
+  {{- end }}
+{{- end }}
```
cpanato, line 135: nit: new line

cpanato (changes requested):
thanks

need to check the helm docs job i think need to update the readme as well and some small nits

haydentherapper:
@cpanato does this look good to you?

cpanato (approved):
thanks

and sorry for the delay

cpanato:
but need to update the docs, can you please take a look?

vipulagarwal:
> but need to update the docs, can you please take a look?

@cpanato the [README](https://github.com/sigstore/helm-charts/pull/744/files#diff-957e39202603703da3b433c7671998188db4dcbea68e2fed85cbf8256f1ee08bR48) was updated
is there anything else missing? would you like the `description` field to be populated?

cvegagimenez:
@cpanato , the tests are failing but I am not able to find the issue. Could you help here?

`{"level":"fatal","ts":1728294947.0237236,"logger":"fallback","caller":"server/main.go:247","msg":"failed to parse /var/run/tuf-secrets/tsa.certchain.pem: unmarshaling certificates: error during PEM decoding","stacktrace":"main.main\n\tgithub.com/sigstore/scaffolding/cmd/tuf/server/main.go:247\nruntime.main\n\truntime/proc.go:272"}`

charts/tuf/templates/deployment.yaml
```diff
@@ -47,40 +47,9 @@ spec:
...
-              items:
-              - key: {{ .Values.secrets.tsa.key }}
-                path: {{ .Values.secrets.tsa.path }}
+          sources:
+          {{- include "tuf.validateSecrets" . }}
     {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
 {{ toYaml .Values.imagePullSecrets | indent 8 }}
     {{- end }}
-    {{- if .Values.deployment.nodeSelector }}
```
vipulagarwal, line 75: why are we removing the `nodeSelector`, `tolerations` and `affinity` features for tuf deployment?

vipulagarwal:
> @cpanato , the tests are failing but I am not able to find the issue. Could you help here?
> 
> ```
> {"level":"fatal","ts":1728294947.0237236,"logger":"fallback","caller":"server/main.go:247","msg":"failed to parse /var/run/tuf-secrets/tsa.certchain.pem: unmarshaling certificates: error during PEM decoding","stacktrace":"main.main\n\tgithub.com/sigstore/scaffolding/cmd/tuf/server/main.go:247\nruntime.main\n\truntime/proc.go:272"}
> ```

Looking at [previous test runs](https://github.com/sigstore/helm-charts/actions/runs/9995092856/job/27626425674#step:10:146), we always had this error but it was somehow still passing the chart installation (incorrect behaviour). The error happens because by default all the secrets are enabled and scaffolding tuf server expects a valid certificate for tsa certificate chain.

The test failing now is the correct behavior, presumable a result of latest scaffolding tuf server image. To fix the actual problem, we could set `create: true` for TSA secret but this enables it by default for everyone using the default values.

charts/tuf/templates/deployment.yaml
```diff
@@ -83,4 +64,4 @@ spec:
     {{- if .Values.deployment.affinity }}
       affinity:
 {{ toYaml .Values.deployment.affinity | indent 8 }}
-    {{- end }}
+    {{- end }}
```
cpanato, line 67: nit: new line here

cvegagimenez: @cpanato any news about this?

cpanato (approved):
thanks

PR #530, opened 2023-05-18
feat(fulcio/add-env): Add additional env variables

saisatishkarra:
## Description of the change

Support GCP credentials for external cloud provider workloads

## Additional Information
Similar changes that were made in TSA
- https://github.com/sigstore/helm-charts/commit/2d4695f29f454a3ac0affb616b4571c819a30578
- https://github.com/sigstore/helm-charts/commit/f120177eb41776463e4d75c7e2b526523705551a

 

## Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). Where applicable, update and bump the versions in any associated umbrella chart
- [x] Variables are documented in the `values.yaml` and added to the README.md. The [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-docs --dry-run` to preview the content.
- [ ] JSON Schema generated.
- [ ] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command.

PR #790, opened 2024-07-24
add additional env variables to fulcio chart

Racer159:
## Description of the change

Support GCP credentials for external cloud provider workloads and add additional env vars for specifying values like `VAULT_ADDR`

This is a repickup of https://github.com/sigstore/helm-charts/pull/530

## Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). Where applicable, update and bump the versions in any associated umbrella chart
- [X] Variables are documented in the `values.yaml` and added to the README.md. The [helm-docs](https://github.com/norwoodj/helm-docs) utility can be used to generate the necessary content. Use `helm-docs --dry-run` to preview the content.
- [X] JSON Schema generated.
- [X] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command.

### DIFF
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 2.9.3
+version: 2.10.0
 appVersion: 1.8.7
 
 keywords:
--- a/charts/fulcio/README.md
+++ b/charts/fulcio/README.md
@@ -2,7 +2,7 @@
 
 <!-- This README.md is generated. Please edit README.md.gotmpl -->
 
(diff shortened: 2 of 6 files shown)


Racer159:
On this though would it make sense to use a pattern more similar to Rekor for extra env vars? https://github.com/sigstore/helm-charts/blob/main/charts/rekor/templates/server/deployment.yaml#L105 - it may be nice in many cases to pull env vars from secrets rather than pass them through Helm values.

Racer159:
(would also like to make a similar change to the `rekor` chart - could do that in this PR or a separate one - currently there _is_ a way to force things to work but it isn't a great solution: https://github.com/defenseunicorns/uds-package-sigstore/blob/main/bundle/uds-bundle.yaml#L52)

cmurphy:
No maintainer looked at this in almost two years. It seemed unfair to ask you to rebase after all this time so I went ahead and rebased for you. If you're still around, please have a look and let me know if it still meets your needs. If you're not around, I'll try to get this merged in the next few days.

Issue #791, opened 2024-07-25
add optional cronJob to copySecretJob to avoid stale secrets.


ianhundere:
**Description**

Since the copySecretJob only runs once (initial deploy), TUF's secrets can get stale. this is not super common since these secrets are not expected to change often, but in the event that they do, TUF quickly loses its source of truth.

Suggestion:
- adding an optional cronJob (e.g. w/ a enable/disable switch, ability to update its schedule, cronJob.schedule, etc). 
- if enabled, the name of the copySecretJob would change to `copySecretJob.name`-immediate and the cronJob would be named `copySecretJob.name`-scheduled.

PR #792, opened 2024-07-26
adds optional cronJob to copySecretJob to avoid stale secrets.

ianhundere:
closes https://github.com/sigstore/helm-charts/issues/791

## Description of the change

This PR adds a cronJob in conjunction w/ the copy secrets job.

- https://github.com/sigstore/helm-charts/issues/791

## Additional Information

I included the secret `delete` permission due to the following error:
`Error from server (Conflict): Operation cannot be fulfilled on secrets "<secret_name>": the object has been modified; please apply your changes to the latest version and try again`

^ When trying to apply changes to secrets, I would receive the patch permission error about the resource version. This happens when the resource version of the secret being patched does not match the current resource version stored (e.g. used to manage optimistic concurrency). So if the secret has been modified since retrieved, the resource version will no longer match, thus preventing any further patch operation(s) to avoid overwriting changes made by other operations.

We could also retrieve the latest version of the secret, and then apply our changes to the latest version, and then patch, but I thought simply deleting / applying was better.

## Checklist

- [x] JSON Schema generated.
```
Linting charts...

------------------------------------------------------------------------------------------------------------------------
 Charts to be processed:
------------------------------------------------------------------------------------------------------------------------
 scaffold => (version: "0.6.62", path: "charts/scaffold")
------------------------------------------------------------------------------------------------------------------------

"sigstore" already exists with the same configuration, skipping
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "sigstore" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 6 charts
Downloading fulcio from repo https://sigstore.github.io/helm-charts
Downloading rekor from repo https://sigstore.github.io/helm-charts
Downloading trillian from repo https://sigstore.github.io/helm-charts
Downloading ctlog from repo https://sigstore.github.io/helm-charts
Downloading tuf from repo https://sigstore.github.io/helm-charts
Downloading tsa from repo https://sigstore.github.io/helm-charts
Deleting outdated charts
Linting chart "scaffold => (version: \"0.6.62\", path: \"charts/scaffold\")"
Checking chart "scaffold => (version: \"0.6.62\", path: \"charts/scaffold\")" for a version bump...
Old chart version: 0.6.61
New chart version: 0.6.62
Chart version ok.
Validating /Users/ianhundere/Projects/helm-charts/charts/scaffold/Chart.yaml...
Validation success! 👍

Linting chart with values file "charts/scaffold/ci/ci-values.yaml"...

==> Linting charts/scaffold
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed

------------------------------------------------------------------------------------------------------------------------
 ✔︎ scaffold => (version: "0.6.62", path: "charts/scaffold")
------------------------------------------------------------------------------------------------------------------------
All charts linted successfully
```

### DIFF
--- a/charts/scaffold/Chart.yaml
+++ b/charts/scaffold/Chart.yaml
@@ -4,7 +4,7 @@ description: Scaffolding the components of the sigstore architecture
 
 type: application
 
-version: 0.6.61
+version: 0.6.62
 keywords:
   - security
   - pki
--- a/charts/scaffold/README.md
+++ b/charts/scaffold/README.md
@@ -49,6 +49,11 @@ helm uninstall [RELEASE_NAME]
 |-----|------|---------|-------------|
 | copySecretJob.affinity | object | `{}` |  |
 | copySecretJob.backoffLimit | int | `6` |  |
+| copySecretJob.copySecretCronJob.backoffLimit | int | `2` |  |
+| copySecretJob.copySecretCronJob.enabled | bool | `false` |  |
+| copySecretJob.copySecretCronJob.failedJobsHistoryLimit | int | `1` |  |
+| copySecretJob.copySecretCronJob.schedule | string | `"*/5 * * * 1-5"` |  |
+| copySecretJob.copySecretCronJob.successfulJobsHistoryLimit | int | `1` |  |
 | copySecretJob.enabled | bool | `false` |  |
 | copySecretJob.imagePullPolicy | string | `"IfNotPresent"` |  |
 | copySecretJob.name | string | `"copy-secrets-job"` |  |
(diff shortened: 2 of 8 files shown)


vipulagarwal:
Hi @ianhundere, this feature can be helpful to automate things. Thank you.

Do we need to add some logic around restarting the `tuf-server` to update the TUF root? On the initial setup, `tuf-server` pod keeps on failing unless these secrets are populated by the `copySecretsJob`. I am unsure if it will pickup the new keys from the updated secrets by the cron job.

ianhundere:
@vipulagarwal i don't believe that is necessary as per:
“When a volume contains data from a Secret, and that Secret is updated, Kubernetes tracks this and updates the data in the volume, using an eventually-consistent approach.
A container using a Secret as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes#using-subpath) volume mount does not receive automated Secret updates.”
^[Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically)

more info here:

[Why Kubernetes secrets take so long to update?](https://ahmet.im/blog/kubernetes-secret-volumes-delay/) 

“The kubelet checks whether the mounted ConfigMap is fresh on every periodic sync. However, it uses its local TTL-based cache for getting the current value of the ConfigMap. As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + TTL of ConfigMaps cache (1 minute by default) in kubelet.”

[ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#mounted-configmaps-are-updated-automatically)

i've done tests w/o running into issues.

ianhundere:
@sabre1041 / @cpanato is this something that would be considered as a contribution?

cheers / thanks!

ianhundere:
@bobcallaway do you think there'd be any interest in something like this?

ianhundere:
@bobcallaway that's a good point, and while we are using external secrets, those external secrets point to the secrets within' each respective service namespace (e.g. trillian-system, rekor-system, fulcio-system, ctlog-system, tsa-system etc), but what happens is if one of those secrets is updated, TUF won't know about it since it runs only once and never again checks to see if its respective secrets are up-to-date unless TUF's copied secrets are removed "and" deploy TUF again (e.g. rerun the copySecretJob).

tl;dr provide a quick/native way to ensure TUF's copied secrets are updated/pointing to their respective source of truth regularly.

bobcallaway:
ah. in sigstore public-good-instance, we are using https://github.com/stakater/Reloader as a "watcher" to bounce deployments when k8s secrets are changed (and they are auto-updated by external-secrets). 

i'm not against the concept here, just trying to reduce the number of moving pieces w/i the helm-charts.

ianhundere:
> i'm not against the concept here, just trying to reduce the number of moving pieces w/i the helm-charts.

i totally understand, less is often more. 😅 

lemme know and i'll rebase to resolve conflicts.

bobcallaway:
ah, i think i see the issue here now (which is the use of jobs & init containers to dynamically create secrets in the helm charts) vs creating those assets a priori and putting them into a secret manager/vault and using external-secrets. 

I think adding this PR is fine given the existing pattern, but outside of ephemeral testing clusters I'd think a more intentional IaC approach would be a better approach for users.

ianhundere:
yep yep / agreed. thanks for the quick response. i'm outta pocket today, but will rebase/push tomorrow morn.

ianhundere:
@bobcallaway just checking-in to see if there's anything else you'd like me to adjust / clarify in the PR.

thanks again for your consideration and all the work on the sigstore project.

bobcallaway (approved):
LGTM, sorry for the delay!

PR #803, opened 2024-07-31
Create helper for fulcio config yaml

javanlacerda:
## Description of the change

I noticed that [this](https://github.com/sigstore/helm-charts/blob/b0a2dc97ee241bf1ab8b01726a48261108610df0/charts/fulcio/templates/fulcio-configmap.yaml#L16) condition was not working properly when I was bumping the fulcio version in scaffold.

Then I managed to move the logic of create the config with yaml format into the helper and setting the indent correctly.

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/fulcio/templates/fulcio-configmap.yaml
+++ b/charts/fulcio/templates/fulcio-configmap.yaml
@@ -10,12 +10,10 @@ data:
 # config's content.
 # If the field format is empty, the default case is consider that is a
 # json or is empty and should use the defaults as defined on the file _helpers.tpl
-{{- if and (eq .Values.config.format "yaml") (.Values.config.contents) -}}
-  {{- with .Values.config.contents }}
+{{- if eq .Values.config.format "yaml"}}
   config.yaml: |-
-    {{ toYaml . | indent 2 }}
-  {{- end }}
-{{- else -}}
+{{ include "fulcio.configmap.yaml" . | indent 4 }}
+{{- else }}
   config.json: |-
 {{ include "fulcio.configmap.contents" . | indent 4 }}
 {{- end }}
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 2.4.0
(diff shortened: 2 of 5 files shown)


charts/fulcio/templates/fulcio-configmap.yaml
```diff
@@ -10,12 +10,10 @@ data:
 # config's content.
 # If the field format is empty, the default case is consider that is a
 # json or is empty and should use the defaults as defined on the file _helpers.tpl
-{{- if and (eq .Values.config.format "yaml") (.Values.config.contents) -}}
-  {{- with .Values.config.contents }}
+{{- if and (eq .Values.config.format "yaml")}}
```
haydentherapper, line 13: Do you need “and”?

haydentherapper (commented):
Just wondering, what was the error you were getting with the previous code?

charts/fulcio/templates/fulcio-configmap.yaml
javanlacerda: No! removing

javanlacerda:
> Just wondering, what was the error you were getting with the previous code?

I've got this one:

```Error: INSTALLATION FAILED: 1 error occurred:
        * ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal object into Go struct field ConfigMap.data of type string```

haydentherapper (approved):
To confirm, this was tested locally with json and yaml configs and was working?

javanlacerda:
> To confirm, this was tested locally with json and yaml configs and was working?

I tested for the default value, that is json. I'm doing a double check here.

javanlacerda:
> > To confirm, this was tested locally with json and yaml configs and was working?
> 
> I tested for the default value, that is json. I'm doing a double check here.

Tested! We're good

javanlacerda:
> > To confirm, this was tested locally with json and yaml configs and was working?
> 
> I tested for the default value, that is json. I'm doing a double check here.

It is 100% tested now, working properly for json and yaml

haydentherapper (approved):

Issue #217, opened 2022-06-17
policy-webhook-certs secrets are not getting created


MageshSrinivasulu:
Can someone please provide details on how these secrets are getting populated? I my case its not showing. webhook-certs are working fine 

**policy-webhook-certs :**

````
kubectl describe secret policy-webhook-certs
Name:         policy-webhook-certs
Namespace:    genesys-system
Labels:       app.kubernetes.io/instance=policy-controller
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=policy-controller
              app.kubernetes.io/version=1.9.0
              control-plane=policy-controller-policy-webhook
              helm.sh/chart=policy-controller-0.1.25
Annotations:  meta.helm.sh/release-name: policy-controller
              meta.helm.sh/release-namespace: genesys-system

Type:  Opaque

Data
====
````

**webhook-certs :**

```
kubectl describe secret webhook-certs
Name:         webhook-certs
Namespace:    genesys-system
Labels:       app.kubernetes.io/instance=policy-controller
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=policy-controller
              app.kubernetes.io/version=1.9.0
              control-plane=policy-controller-webhook
              helm.sh/chart=policy-controller-0.1.25
Annotations:  meta.helm.sh/release-name: policy-controller
              meta.helm.sh/release-namespace: genesys-system

Type:  Opaque

Data
====
server-cert.pem:  794 bytes
server-key.pem:   241 bytes
ca-cert.pem:      855 bytes
```

hectorj2f:
Those are created during the installation. It is done using knative. I tried increasing the delay of the probes, but it seems it is happening. Have you tried the latest char version?

hectorj2f:
It is done using this functionality  https://knative.dev/docs/serving/using-auto-tls/

MageshSrinivasulu:
> Those are created during the installation. It is done using knative. I tried increasing the delay of the probes, but it seems it is happening. Have you tried the latest char version?

@hectorj2f  Yes I am using the latest available chart only. Is the delay configurable in the helm some how ?

hectorj2f:
The certificates are not created because the leases (kubectl get leases) are not cleaned up for the webhooks. I managed to reproduce it by installing and deleting the chart.

I feel the solution could be one of these three:
* delete the namespace where you installed the policy-controller chart when uninstalling it, if possible.
* delete the leases before every re-installation.
[...]

PR #817, opened 2024-08-30
Leases cleanup add security context

ybelMekk:
## Description of the change

securityContxect to cleanup-leases job

* In a cluster with https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/ or other similar best practice policies, the lease job does not clean up, as it fails to be created, output similar errors:

`run-as-non-root: 'validation error: Running as root is not allowed. Either the field..`

This also relates to issue listed, leases dont get cleaned up the application fails to start if re-deployed. 

https://github.com/sigstore/helm-charts/issues/217#issuecomment-1160155615

## Additional Information

 

Add podSecurity context and security context to values.yaml and corresponding cleanup-lease job.

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/policy-controller/values.yaml
+++ b/charts/policy-controller/values.yaml
@@ -72,6 +72,18 @@ leasescleanup:
     repository: cgr.dev/chainguard/kubectl
     version: latest-dev
     pullPolicy: IfNotPresent
+  securityContext:
+    enabled: false
+    runAsUser: 1000
+    runAsNonRoot: true
+  podSecurityContext:
+    enabled: true
+    allowPrivilegeEscalation: false
+    readOnlyRootFilesystem: true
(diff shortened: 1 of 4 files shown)


charts/policy-controller/values.yaml
```diff
@@ -72,6 +72,18 @@ leasescleanup:
     repository: cgr.dev/chainguard/kubectl
     version: latest-dev
     pullPolicy: IfNotPresent
+  securityContext:
+    enabled: false
+    runAsUser: 1000
+    runAsNonRoot: true
+  podSecurityContext:
```
cmurphy, line 79: It looks like this has been added since you proposed this. Does that address the issue or is the other security context needed as well?

PR #859, opened 2024-10-31
handle extraArgs consistently in ctlog chart as in other charts

bobcallaway:
ctlog chart expected extraArgs as a map instead of a list of strings; this makes it consistent with others

### DIFF
--- a/charts/ctlog/templates/_helpers.tpl
+++ b/charts/ctlog/templates/_helpers.tpl
@@ -110,15 +110,9 @@ Server Arguments
 - {{ printf "--metrics_endpoint=0.0.0.0:%d" (.Values.server.portHTTPMetrics | int) | quote }}
 - "--log_config=/ctfe-keys/config"
 - "--alsologtostderr"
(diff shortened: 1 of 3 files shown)


charts/ctlog/templates/_helpers.tpl
```diff
@@ -110,15 +110,9 @@ Server Arguments
...
-{{- if .Values.server.extraArgs -}}
-{{- range $key, $value := .Values.server.extraArgs }}
-{{- if $value }}
-- {{ printf "%v=%v" $key $value | quote }}
-{{- else }}
-- {{ printf $key | quote }}
-{{- end }}
-{{- end }}
-{{- end -}}
+{{- range .Values.server.extraArgs }}
```
haydentherapper, line 113: To confirm, it doesnt need an if wrapping around it?

bobcallaway: no because we have other elements in the list (lines 109-112) already.

PR #862, opened 2024-11-01
Replace chainguard netcat image with public docker.io image

priyawadhwa:
Chainguard is removing the netcat image from the free tier, so replace it with a public image from dockerhub. this one seemed the most popular of all the options (~5M pulls)

### DIFF
--- a/charts/trillian/README.md
+++ b/charts/trillian/README.md
@@ -2,7 +2,7 @@
 
 <!-- This README.md is generated. Please edit README.md.gotmpl -->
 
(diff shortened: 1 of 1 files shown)


haydentherapper:
Given our usage is just [polling](https://github.com/sigstore/helm-charts/blob/9f5d055ecdad62acf1ff39e650f4d95f204d102b/charts/trillian/templates/trillian-log-server/deployment.yaml#L43), I wonder if there's another utility, wget or curl, in an existing container dependency we can use. Otherwise LGTM, just need to regen docs

PR #943, opened 2025-04-01
TSA: remove hardcoded credentials file

### LINKED ISSUE #940: tsa chart hard codes GOOGLE_APPLICATION_CREDENTIALS
`GOOGLE_APPLICATION_CREDENTIALS: /etc/tsa-config/cloud_credentials`

This seems to be problematic (if credentials file is not needed, tink client seems to fail trying to read an empty file): details are a little unclear still, but I'm filing so I remember to update the chart once smoke has cleared.

jku:
There's nothing urgent here but it allows removing `GOOGLE_APPLICATION_CREDENTIALS` from public good instance scaffold.

* tsa
   *  Don't hard code GOOGLE_APPLICATION_CREDENTIALS. I left the env variable in there since I think it may be useful in other use cases
   * remove a bit of repetition in tsa-deployment
   * bump version
* README: Add missing charts

Fixes #940

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,4 +64,6 @@ NOTE: The public key that was used to sign a particular chart may not be identic
 * [scaffold](charts/scaffold)
 * [sigstore-prober](charts/sigstore-prober)
 * [trillian](charts/trillian)
+* [tsa](charts/tsa)
+* [tuf](charts/tuf)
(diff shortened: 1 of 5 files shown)


haydentherapper:
Thanks, will we want a scaffold chart bump after this for any cleanup in PGI?

PR #945, opened 2025-04-04
Add rekor-tiles chart

cmurphy:
## Description of the change

Split into two commits - the first is the result of running `helm create rekor-tiles`, the second shows the customizations made to the boilerplate to support rekor-tiles. This seemed more straightforward than copying the old chart and trying to modernize it.

Partial https://github.com/sigstore/rekor-tiles/issues/41

## Additional Information

 

This adds support for GCP [workload identity](https://cloud.google.com/iam/docs/workload-identity-federation) which is the [recommended](https://cloud.google.com/docs/authentication?#auth-decision-tree) authentication method for a GKE application. This seems to be the first chart using this type of application identity, I can remove this to be more in line with the way the other charts are written if needed.

## Checklist

- [x] JSON Schema generated.


cmurphy:
@haydentherapper @bobcallaway

charts/rekor-tiles/rekor-tiles/templates/ingress.yaml
```diff
@@ -0,0 +1,81 @@
...
+    {{- range .Values.ingress.tls }}
+    - hosts:
+      {{- range .hosts }}
+      - {{ . | quote }}
+      {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+{{- end }}
+{{- end -}}
+{{- range .Values.ingresses }}
```
bobcallaway, line 34: We should collapse this template into a single one for `ingress`; the only reason there are both `ingress` and `ingresses` are when we made a transition from NGINX to GCE load balancers in pgi

charts/rekor-tiles/rekor-tiles/templates/deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ template "rekor-tiles.image" .Values.image }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: GOOGLE_APPLICATION_CREDENTIALS
```
bobcallaway, line 38: I would have expected the GCP clients to read this from metadata instead of explicitly having to pass this through; does this enable local testing?

charts/rekor-tiles/rekor-tiles/templates/deployment.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
bobcallaway, line 3: we should support specifying annotations on the deployment

charts/rekor-tiles/rekor-tiles/templates/_helpers.tpl
```diff
@@ -0,0 +1,182 @@
...
+{{- end }}
+{{- end }}
+{{- if .Values.signer.kms }}
+{{- if .Values.signer.kms.hash }}
+- {{ printf "--signer-kmshash=%s" .Values.signer.kms.hash | quote }}
+{{- end }}
+- {{ printf "--signer-kmskey=%s" .Values.signer.kms.key | quote }}
+{{- if .Values.signer.tink }}
+- {{ printf "--signer-tink-kek-uri=%s" .Values.signer.tink.kekURI | quote }}
+- "--signer-tink-keyset-path=/etc/tink-config/tink.keyset.enc
```
bobcallaway, line 144: it would be helpful to also be able to have a caller specify `extraArgs` which get appended without having to update the helm chart

cmurphy: done

charts/rekor-tiles/rekor-tiles/templates/deployment.yaml
cmurphy: done

charts/rekor-tiles/rekor-tiles/templates/deployment.yaml
cmurphy: I was overcomplicating it by not using GKE. Removed all this.

charts/rekor-tiles/rekor-tiles/templates/ingress.yaml
cmurphy: Done

charts/rekor-tiles/rekor-tiles/templates/deployment.yaml
haydentherapper: BTW, if you have any suggestions for how to improve the timestamp authority helm chart, we should update that as well. I required GOOGLE_APPLICATION_CREDENTIALS be copied into the image but at the time I didn't really know how else to do it.

charts/rekor-tiles/templates/_helpers.tpl
```diff
@@ -0,0 +1,185 @@
...
+- {{ printf "--checkpoint-interval=%s" .Values.tesseraLifecycle.checkpointInterval | quote }}
+{{- end }}
+{{- if .Values.tesseraLifecycle.pushbackMaxOutstanding }}
+- {{ printf "--pushback-max-outstanding=%d" (.Values.tesseraLifecycle.pushbackMaxOutstanding | int) | quote }}
+{{- end }}
+{{- end }}
+{{- if .Values.clientSigningAlgorithms }}
+- {{ printf "--client-signing-algorithms=%s" .Values.clientSigningAlgorithms | quote }}
+{{- end }}
+- {{ printf "--gcp-bucket=%s" .Values.gcp.bucket | quote }}
```
haydentherapper, line 98: Should we check `if .Values.gcp` for a future iteration of supporting other backends?

charts/rekor-tiles/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.signer.tink }}
```
haydentherapper, line 1: JW, why a configmap for tink and not for the file signer?

haydentherapper: Oh I think I answered my question, it's a secret rather than a configmap, yea?

cmurphy: Trying to mirror what TSA does https://github.com/sigstore/helm-charts/blob/main/charts/tsa/templates/tsa-configmap.yaml

Rekor v1 doesn't seem to have tink support in its helm chart so I didn't have a better example to go from.

No it's not a secret, if it should be I can change it (that's another ExternalSecret we'd need to manage).

charts/rekor-tiles/templates/_helpers.tpl
cmurphy: Right now there's no other option so it is a hard requirement, I'd say we can be more flexible when we have other backends to support

charts/rekor-tiles/templates/configmap.yaml
cmurphy: Right, it will be a private key and possibly also a password. I only added the secret at all to make CI happy since it depended on the secret existing, otherwise this should be created out of band in a more secure way.

cmurphy:
Removed rekor-tiles from the CT config so the CI won't try to deploy it. Like some other charts, it has a hard dependency on external resources and can't be deployed successfully without real configuration. In a real deployment we would want it to fail if it can't reach those resources, so I don't see a good reason to try to add in a workaround.

haydentherapper (approved):
Great work!

charts/rekor-tiles/values.yaml
```diff
@@ -0,0 +1,137 @@
...
+  #     mountPath: /pki
+  #     mountSubPath: signer.pem
+  # kms:
+  #   hash: sha256
+  #   key: gcpkms://keyname
+  # tink:
+  #   kekURI: gcp-kms://keyname
+  #   configMapName: tink-config
+  #   keyset: ""
+extraArgs: []
```
bobcallaway, line 137: perhaps a nit but most (if not all) charts in this repo nest all server arguments under `server` instead of having them as top-level... for consistency it might be nice to do that but certainly open to comments

cmurphy: I veered away from that on purpose but I don't feel that strongly about it. The reason rekor does it is that it has multiple deployments and each main section (`server`, `mysql`, and `redis`) are configuring that specific deployment https://github.com/sigstore/helm-charts/tree/main/charts/rekor/templates

Here we're just configuring one thing. I guess I could nest them preemptively in case we add more stuff?

haydentherapper: We will eventually support mysql, so I could see us using the same pattern.

cmurphy: Done

haydentherapper (approved):

bobcallaway:
last thing i see: all the other charts explicitly specify namespace everywhere as well as a `namespace.create` option. Should we repeat that in here? 

otherwise LGTM

cmurphy:
Helm can create the namespace with `--create-namespace` and this will all be in one namespace, so I think it's not needed.

bobcallaway:
> Helm can create the namespace with `--create-namespace` and this will all be in one namespace, so I think it's not needed.

right... but i'm not sure whether ArgoCD will pull that through in how we have it in pgi

haydentherapper:
I'm not well-versed in helm/argocd, but I think the lack of an explicit namespace could cause argocd to think there's constantly a diff. Can we add it for uniformity/just in case?

haydentherapper:
I'm not sure the history behind it. I can't see why that wouldn't be safe, with including a default name.

haydentherapper (approved):

haydentherapper:
For future infra work, I wonder how much of this we can reuse for the static-ct deployment. Off the top of my head, the primary differences would be the server args and how the signing key is mounted (No Tink support, they're loading in the signing key from GCP Secrets).

cmurphy:
> I'm wondering if this can be turned into a template for any tile-backed log, using that template for the rekor-tiles chart and a future static-ct chart.

Couple of thoughts here.

For one, I'm not much of a helm wizard (I do not enjoy templates) but in my experience duplication is pretty common in helm charts, and trying to be clever by abstracting away common parts would be an anti-pattern that might introduce more confusion than it's worth.

For another, this helm chart really has little to do with tiles at all. It's a set of resources that deploy a particular server image. The only elements it has in common with another tiles deployment would actually be the server arguments pointing to the storage backend. Otherwise, every resource in it is common to every other chart in our repo - a deployment, a service account, a namespace, an ingress. The only thing unique to this chart is the tink configmap, which as you said would not be useful for other CT.

cmurphy:
@bobcallaway do you see any other changes needed for this PR?

bobcallaway (approved):

PR #970, opened 2025-04-28
Remove rekor-tiles Ingress in favor of NEGs

cmurphy:
The rekor-tiles service does not directly enable its own read path, but instead relies on external routing to a GCS bucket. The GCP Load Balancer created by a Kubernetes Ingress resource is not capable of routing to anything outside of Kubernetes, so this load balancer must be managed in Terraform instead.

This change removes the rekor-tiles Ingressa nd supporting resources entirely, and instead adds a Network Endpoint Group annotation to the Service resource, which will create the Network Endpoint Group[1] in GCP, which Terraform can then use to route to the cluster.

[1] https://cloud.google.com/load-balancing/docs/negs/zonal-neg-concepts

Partial https://github.com/sigstore/rekor-tiles/issues/239
Depends on https://github.com/sigstore/terraform-modules/pull/24

## Checklist

- [ ] JSON Schema generated.


charts/rekor-tiles/templates/service.yaml
```diff
@@ -5,10 +5,12 @@ metadata:
   namespace: {{ .Values.namespace.name }}
   labels:
     {{- include "rekor-tiles.labels" . | nindent 4 }}
+  annotations:
+    cloud.google.com/neg: '{"exposed_ports": {"80":{"name": "{{ .Values.neg.name }}"}}}'
```
haydentherapper, line 9: `80` needs to match what's set from `.Values.service.ports[0].port`, correct? Do we need to pull the port num from the template?

Do we also need to specify an ordering for the ports, that the first is for the NEG, second for gRPC, third for metrics (basing this on the current ordering)?

haydentherapper (commented):
Wow, this greatly simplifies things!

haydentherapper (approved):
Fantastic work!

PR #990, opened 2025-05-24
Trillian: fix log signer forceMaster always true

ajh-:
## Description of the change
`forceMaster` value stays `true` even when set to `false` in Trillian's values.yaml

## Additional Information
Workaround is to include `--force_master=false` in `extraArgs`

 

## Checklist

- [ ] JSON Schema generated.


ajh-:
I definitely took something simple and made it a lot more complicated.

bobcallaway (approved):

PR #1042, opened 2025-09-10
rekor-tiles: Set port names in deployment spec

jku:
Also use descriptive name for the metrics port.

The purpose is to test the theory that PodMonitoring requires port names to fully work: "The container metadata label is only populated if the port is referenced by name because port numbers are not unique across containers."

@cmurphy do you think this is worth testing out (requires also updating the relevant podmonitoring yaml to `port: metrics` )?

### DIFF
--- a/charts/rekor-tiles/Chart.yaml
+++ b/charts/rekor-tiles/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rekor-tiles
 description: Part of the sigstore project, Rekor v2 (Rekor on tiles) is a signature transparency log
 type: application
-version: 0.2.17
+version: 0.2.18
 appVersion: "0.1.11"
 keywords:
   - security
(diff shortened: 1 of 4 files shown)


jku:
> "The container metadata label is only populated if the port is referenced by name because port numbers are not unique across containers."

This seems to really only refer to `container` label specifically and not any others so it likely does not help with the main issue in https://github.com/sigstore/rekor-tiles/issues/452

cmurphy:
I'm not convinced this will help with the issue but it's still nice to have a clear name on the port. I would also wonder why we don't have this problem with other services (or do we?).

You'll also want to change the port reference in https://github.com/sigstore/public-good-instance/blob/main/argocd/utilities/manifest/staging/prometheus-monitoring/rekor-tiles.yaml to actually test the theory.

cmurphy (approved):

PR #1063, opened 2025-11-04
Remove duplicate service type key from service template

jhart0:
## Description of the change

The service template contains a duplicate type attribute. This breaks tooling that uses server-side apply such as Flux V2 due to the duplicate key. Tooling using client-side apply (e.g. helm cli) are not affected.

## Additional Information

This changes takes the value from the supplied values.yaml, which defaults to ClusterIP anyway.

## Checklist

- [x] JSON Schema generated.


haydentherapper:
Added this change to https://github.com/sigstore/helm-charts/pull/1069, thanks!

PR #1098, opened 2025-12-04
Add ctlog-tiles chart

cmurphy:
Add a chart named ctlog-tiles to support TesseraCT as an alternative CT log for Fulcio. The POSIX and GCP backends are supported.

This chart has a unique way of crafting the image string based on which personality (posix or gcp) is chosen. This lets values.yaml/Chart.yaml still be the source of truth for the image version, which is consistent with how Sigstore charts are used in the public good instance (as opposed to overriding the image version in the infrastructure values.yaml).

This is closely based on the rekor-tiles chart. It creates NEGs but no explicit Ingress gateway, since a multi-purpose load balancer needs to be created externally to link the service to the backend bucket.

Relates to https://github.com/sigstore/rekor-tiles/issues/73

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/ctlog-tiles/Chart.yaml
+++ b/charts/ctlog-tiles/Chart.yaml
@@ -1,0 +1,21 @@
+apiVersion: v2
+name: ctlog-tiles
+description: Tiles-based certificate log (TesseraCT)
+type: application
+version: 0.1.0
+appVersion: v0.1.1
+keywords:
+  - security
+  - certificate transparency
+home: https://sigstore.dev
+sources:
+  - https://github.com/transparency-dev/tesseract
+maintainers:
+  - name: The Sigstore Authors
+annotations:
+  artifacthub.io/license: Apache-2.0
+  artifacthub.io/images: |
+    - name: tesseract-posix
+      image: ghcr.io/sigstore/scaffolding/tesseract/posix:v0.1.1@sha256:767c2dbe4a83be9243246564244aa381322f08f2a864371f62b880c465033815
+    - name: tesseract-gcp
+      image: ghcr.io/sigstore/scaffolding/tesseract/gcp:v0.1.1@sha256:17d3788b736de01ab0586359626c01789910a6e2981e7b90014fbda8689f5017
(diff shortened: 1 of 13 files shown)


charts/ctlog-tiles/README.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+## Using the POSIX backend
+
+In values.yaml, set `image.flavor` to `"posix"` (default).
+
+Create a secret for the private key:
+
+```
+kubectl -n ctlog-tiles-system create secret generic ctlog-signing-key --from-file=signing-key=/path/to/privkey.pem
```
haydentherapper: Can we give instructions on generating the private key too, just an openssl incantation?

charts/ctlog-tiles/templates/deployment.yaml
```diff
@@ -0,0 +1,128 @@
...
+      serviceAccountName: {{ include "ctlog-tiles.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ template "ctlog-tiles.image" .Values.image }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          #env:
```
haydentherapper: Why is this section commented out?

charts/ctlog-tiles/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: ctlog-tiles
+description: Tiles-based certificate log (TesseraCT)
+type: application
+version: 0.1.0
+appVersion: v0.1.1-0.20251112113307-52b154e44576
```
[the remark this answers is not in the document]
cmurphy: Thanks! https://github.com/sigstore/scaffolding/pull/1835

PR #1102, opened 2025-12-11
Use less confusing name for security contexts

jonkerj:
## Description of the change

The pod-level security context of policy controller's webhook was configured using `webhook.securityContext`, while the container-level security context was using `webhook.podSecurityContext`. The former is debatable, the latter is just plain confusing.

This change uses explicit names for both contexts, removing all confusion.

n/a, if needed I will create one

## Additional Information

This is probably a breaking change, as users overriding security context need to change their values.

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/policy-controller/values.yaml
+++ b/charts/policy-controller/values.yaml
@@ -38,12 +38,12 @@ webhook:
     requests:
       cpu: 100m
       memory: 128Mi
-  securityContext:
+  podSecurityContext:
     enabled: false
     runAsUser: 65532
   failurePolicy: Fail
   podAnnotations: {}
-  podSecurityContext:
+  containerSecurityContext:
     enabled: true
     allowPrivilegeEscalation: false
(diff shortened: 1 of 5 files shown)


charts/policy-controller/values.yaml
```diff
@@ -38,12 +38,12 @@ webhook:
     requests:
       cpu: 100m
       memory: 128Mi
-  securityContext:
+  podSecurityContext:
```
cmurphy, line 41: I notice other charts use `containerSecurityContext` and `securityContext`, it's mainly only this chart that referred to `podSecurityContext` before. Could we make this just `securityContext` to be consistent?

That would also give the opportunity to add some backwards-compatible logic, where the template could use `podSecurityContext` in place of `containerSecurityContext` in case the user had it explicitly defined in their custom values.yaml.

PR #1127, opened 2026-01-15
Add init container for Fulcio

cmurphy:
Do not let the Fulcio service start if the CT log URL is wrong. This will prevent a rollout from "successfully" completing if the CT log cannot be reached or the URL was incorrectly provided, allowing it to stay on the old revision while the operator corrects the error.

This is a test to prevent https://github.com/sigstore/public-good-instance/pull/3515 from happening.

## Checklist

- [ ] JSON Schema generated.

### DIFF
--- a/charts/fulcio/Chart.yaml
+++ b/charts/fulcio/Chart.yaml
@@ -5,7 +5,7 @@ description: |
 
 type: application
 
-version: 2.7.2
+version: 2.7.3
 appVersion: 1.8.5
 
 keywords:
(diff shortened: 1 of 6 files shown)


loosebazooka:
We can revert this for now. And come back to it.

Hrmm there were some other prs since. I'll have to follow the path of changes

mattmoor:
@loosebazooka yeah I think adding CI to deploy this would be great.

We can try looking into pinning things, but wanted to flag it here.

loosebazooka:
gotcha, yeah I'll look into adding the test here. And maybe we can roll forward a new scaffold chart with the old fulcio chart ref if you can't pin.

mattmoor:
Just to close the loop here, pinning back seems to have unblocked our downstream CI.  Thanks, @loosebazooka

PR #1174, opened 2026-05-27
Add persistent storage support for Redis

rseleven:
Add PVC configuration for Redis with support for:
- ReadWriteOnce and ReadWriteMany access modes
- Custom StorageClass selection
- Using an existing PVC
- Automatic deployment strategy selection based on access mode (Recreate for RWO, default RollingUpdate for RWX)

When persistence is disabled, emptyDir is used as before.

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/rekor/templates/redis/deployment.yaml
+++ b/charts/rekor/templates/redis/deployment.yaml
@@ -18,11 +18,10 @@ spec:
     matchLabels:
       {{- include "rekor.redis.matchLabels" . | nindent 6 }}
   replicas: {{ .Values.redis.replicaCount }}
-  {{- if .Values.redis.strategy }}
+  {{- if and .Values.redis.persistence.enabled (eq .Values.redis.persistence.accessMode "ReadWriteOnce") }}
   strategy:
-{{ toYaml .Values.redis.strategy | trim | indent 4 }}
-    {{ if eq .Values.redis.strategy.type "Recreate" }}rollingUpdate: null{{ end }}
-{{- end }}
+    type: Recreate
+  {{- end }}
   template:
     metadata:
     {{- if .Values.redis.podAnnotations }}
@@ -96,5 +95,14 @@ spec:
     {{- end }}
       volumes:
         - name: storage
+        {{- if .Values.redis.persistence.enabled }}
+          persistentVolumeClaim:
+            {{- if .Values.redis.persistence.existingClaim }}
+            claimName: {{ .Values.redis.persistence.existingClaim }}
+            {{- else }}
+            claimName: {{ template "rekor.redis.fullname" . }}
+            {{- end }}
+        {{- else }}
           emptyDir: {}
+        {{- end }}
 {{- end }}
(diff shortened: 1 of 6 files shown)


charts/rekor/templates/redis/deployment.yaml
```diff
@@ -18,11 +18,10 @@ spec:
     matchLabels:
       {{- include "rekor.redis.matchLabels" . | nindent 6 }}
   replicas: {{ .Values.redis.replicaCount }}
-  {{- if .Values.redis.strategy }}
+  {{- if and .Values.redis.persistence.enabled (eq .Values.redis.persistence.accessMode "ReadWriteOnce") }}
   strategy:
-{{ toYaml .Values.redis.strategy | trim | indent 4 }}
-    {{ if eq .Values.redis.strategy.type "Recreate" }}rollingUpdate: null{{ end }}
-{{- end }}
+    type: Recreate
```
cmurphy, line 23: This isn't backwards compatible for anyone who had a custom strategy set, do you have a justification for that?

charts/rekor/values.yaml
```diff
@@ -30,6 +30,13 @@ redis:
     # -- 6.2.17-alpine3.21
     version: "sha256:148bb5411c184abd288d9aaed139c98123eeb8824c5d3fce03cf721db58066d8"
   resources: {}
+  persistence:
+    enabled: true
```
cmurphy: Set this default to false so that users aren't automatically opted in to the new storage mode

charts/rekor/templates/redis/deployment.yaml
rseleven: The Helm chart has several issues:
1. Redis is deployed as a regular Deployment rather than a StatefulSet, which precludes setting up replication between instances. Each pod runs in standalone mode with its own independent storage. Therefore, with replicaCount > 1, application requests will be routed to different, unsynchronized Redis instances, which inevitably leads to logical errors and data inconsistency. The only correct solution is a single replica.
2. The .spec.strategy.type field can be either Recreate or RollingUpdate.
3. A PVC in ReadWriteOnce mode can only be mounted to a single node. With the RollingUpdate strategy, Kubernetes will attempt to start a new pod before stopping the old one, but will be unable to mount the same PVC to the new pod on a different node — the deployment will hang waiting. Therefore, the only correct combination is: ReadWriteOnce + strategy.type: Recreate.

charts/rekor/values.yaml
rseleven: Ok

charts/rekor/values.yaml
```diff
@@ -30,6 +30,13 @@ redis:
     # -- 6.2.17-alpine3.21
     version: "sha256:148bb5411c184abd288d9aaed139c98123eeb8824c5d3fce03cf721db58066d8"
   resources: {}
+  persistence:
+    enabled: false
```
rseleven, line 34: enabled: false

charts/rekor/templates/redis/deployment.yaml
cmurphy: Thanks for the explanation

cmurphy (approved):

Issue #1182, opened 2026-05-29
policy-controller: chart provides duplicated resources for webhook configuration


ttryf:
**Description**

Helm chart for `policy-controller` ships two resources for `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` webhook configuration.

#### How to reproduce:

- Deploy chart using the default values.

The following validating and mutating webhook configurations are in the cluster after the installation.

```sh
❯ kubectl get validatingwebhookconfigurations
NAME                                         WEBHOOKS   AGE
policy.sigstore.dev                          1          2m9s
validating.clusterimagepolicy.sigstore.dev   1          2m9s
```
and

```
❯ kubectl get mutatingwebhookconfigurations.admissionregistration.k8s.io
NAME                                         WEBHOOKS   AGE
defaulting.clusterimagepolicy.sigstore.dev   1          72s
policy.sigstore.dev                          1          72s
```

The issue becomes more apparent when additional Helm values are in place, as an example when `failurePolicy` is set:

```yaml
❯ cat values-kind.yaml
webhook:
  failurePolicy: Ignore
  configData:
    no-match-policy: "warn"
  namespaceSelector:
    matchExpressions: []
```
Based on the above the same resources are created in the k8s cluster but `failurePolicy` isn't respected in one of them. Rendered templates:

*MutatingWebhookConfiguration*
```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    meta.helm.sh/release-name: policy-controller
    meta.helm.sh/release-namespace: cosign-system
  creationTimestamp: "2026-05-28T18:54:03Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
  name: defaulting.clusterimagepolicy.sigstore.dev
  resourceVersion: "1629"
  uid: 8fce9f58-6dbc-407c-aa4e-b740f076810f
webhooks:
- admissionReviewVersions:
  - v1
  clientConfig:
    service:
      name: webhook
      namespace: cosign-system
      port: 443
  failurePolicy: Fail # policy is hardcoded in the template
  matchPolicy: Equivalent
  name: defaulting.clusterimagepolicy.sigstore.dev
  namespaceSelector: {}
  objectSelector: {}
  reinvocationPolicy: Never # gets the default value
  sideEffects: None
  timeoutSeconds: 10
```
and

```yaml
[...]

PR #1206, opened 2026-07-31
fix(policy-controller): Define failurePolicy variable for cosign webhook

ttryf:
## Description of the change

Introduce `failurePolicy` variable for the cosign webhook to parameterise the default action (from the hardcoded `Fail` value).

Relates to #1182 issue

## Checklist

- [x] JSON Schema generated.

### DIFF
--- a/charts/policy-controller/Chart.yaml
+++ b/charts/policy-controller/Chart.yaml
@@ -8,7 +8,7 @@ sources:
 type: application
 
 name: policy-controller
-version: 0.10.7
+version: 0.11.0
 appVersion: 0.13.1
 
 maintainers:
--- a/charts/policy-controller/Chart.yaml
+++ b/charts/policy-controller/Chart.yaml
@@ -8,7 +8,7 @@ sources:
 type: application
 
 name: policy-controller
-version: 0.10.6
+version: 0.11.0
 appVersion: 0.13.1
 
 maintainers:
--- a/charts/policy-controller/README.md
+++ b/charts/policy-controller/README.md
@@ -2,7 +2,7 @@
 
 <!-- This README.md is generated. Please edit README.md.gotmpl -->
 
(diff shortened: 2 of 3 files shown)


charts/policy-controller/Chart.yaml
```diff
@@ -8,7 +8,7 @@ sources:
 type: application
 
 name: policy-controller
-version: 0.10.6
+version: 0.10.7
```
cmurphy: Can you bump this to 0.11.0 since this is adding a feature?

cmurphy (commented):
Based on https://github.com/sigstore/helm-charts/pull/553 it seems like there was an intent at some point to unify these two webhooks and not have separate configuration for them. Could the new `cosign.failurePolicy` be removed and could the other webhoook just use `webhook.failurePolicy`?

ttryf:
> Based on #553 it seems like there was an intent at some point to unify these two webhooks and not have separate configuration for them. Could the new `cosign.failurePolicy` be removed and could the other webhoook just use `webhook.failurePolicy`?

Thanks for the context @cmurphy, this is perfectly fine and it was my initial implementation (before submitting this PR).

One thing I noticed while working on this and calling it out for visibility (my rationale for separate configuration for `failurePolicy`):

The webhook is used to make calls to Fulcio/Rekor/TUF etc. and could fail independently of pod health, compared to ClusterImagePolicy webhook that does local validation - no dependencies. Keeping them separately configurable, reduces the blast radius (during an outage or service unavailability) of each functionality. Not blocking this PR either way, just flagging in case it changes your view on #553.

This is based on the three failure scenarios I considered:

1. Pod is fully down/unavailable: both cases fail regardless of the defined `failurePolicy` value.
2. Pod is healthy, external 3rd party verification is degraded - cosign webhook fails per request while ClusterImagePolicy (CIP) remains unaffected. This is the case where the asymmetry adds value with the `Ignore` case to mitigate a failure that is out of someone's control and it doesn't relate with the correctness of the policy object I'm admitting in the cluster.
3. An object fails admission and gets Ignored/Failed based on the CIP configuration.

charts/policy-controller/Chart.yaml
ttryf: Sure, I rebased my PR and bumped to `0.11.0` version.

PR #1220, opened 2026-08-28
pass actual (fake) keys/certs, add readiness/liveness probes to tuf chart

bobcallaway:
The chart was failing to start because the values provided for the various keys and certificates were just junk strings. That demonstrated itself as an intermittent failure in the CI runs where sometimes the container would show ready even though it was crashing in the background, and other times it would show failed. This hopefully cleans up those intermittent failures.


loosebazooka (approved):

charts/tuf/README.md
```diff
@@ -23,11 +23,25 @@ A framework for securing software update systems - the scaffolding implementatio
 | deployment.affinity | object | `{}` |  |
 | deployment.containerSecurityContext | object | `{}` |  |
 | deployment.imagePullPolicy | string | `"IfNotPresent"` |  |
+| deployment.livenessProbe.failureThreshold | int | `3` |  |
```
loosebazooka, line 26: this seems slightly out of place in this PR?

bobcallaway: Without the readiness probe, it doesn't close the race condition on whether the container actually successfully started or not.

PR #1219, opened 2026-08-27
global image registry override - pt1

bobcallaway:
uses https://helm.sh/docs/chart_template_guide/subcharts_and_globals/ to specify when you want to pull images through a proxy

supports 2 modes:
```
  # Option 1: Multi-registry mapping (e.g., Artifact Registry)
  global:
    imageRegistry:
      ghcr.io: us-docker.pkg.dev/proxy/ghcr
      docker.io: us-docker.pkg.dev/proxy/docker-hub
      quay.io: us-docker.pkg.dev/proxy/quay
      gcr.io: us-docker.pkg.dev/proxy/gcr

  # Option 2: Single unified proxy (e.g., Harbor / Artifactory)
  global:
    imageRegistry: harbor.internal.corp/sigstore-cache
```

PR #1223, opened 2026-08-28
global image registry override - pt2

bobcallaway:
same as #1219 but for fulcio/rekor charts

### DIFF
--- a/charts/rekor/templates/_helpers.tpl
+++ b/charts/rekor/templates/_helpers.tpl
@@ -350,9 +350,9 @@ Create the image path for the passed in image field
   {{- end -}}
 {{- end -}}
 {{- if eq (substr 0 7 $tag) "sha256:" -}}
-{{- printf "%s/%s@%s" $registryName $repositoryName $tag -}}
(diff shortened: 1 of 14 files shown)


charts/rekor/templates/_helpers.tpl
```diff
@@ -325,10 +325,34 @@ Create the name of the service account to use for the Trillian Log Signer compon
...
+    {{- end -}}
+  {{- end -}}
+  {{- if kindIs "map" .global.imageRegistries -}}
+    {{- if hasKey .global.imageRegistries $registryName -}}
+      {{- $registryName = index .global.imageRegistries $registryName -}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+{{- if eq (substr 0 7 $tag) "sha256:" -}}
+{{- printf "%s/%s@%s" $registryName $repositoryName $tag -}}
```
loosebazooka: Feels like this could be indented for readability

loosebazooka (approved):

